PR #12, opened 2019-04-26
Follow-up to  #8 to include actual ability to include unpublished objects

NiklasBr:
While the previous PR only added the field as a return value (which only ever returned `true`) this allows us to get the publication status like this to include both published and unpublished objects:
```
  {
  getProductListing(published: false) {
    edges {
      node {
        id
        name
        sku
        published
...
```


weisswurstkanone:
Thx!

Issue #13, opened 2019-04-29
Thumbnail Transformer is not aware of media queries


NiklasBr:
## Steps to reproduce

1. Create a new *Image Thumbnail*
2. Define at least one breakpoint.
3. Run a client query against the thumbnail field.

## Expected result

To either be able to get an entire `<img srcset...>` tag or an array with data objects representing the different image objects (which may include the different paths).

## Actual result

I get the path to one of the thumbnails only.

weisswurstkanone:
@NiklasBr can this be closed ?

NiklasBr:
I tried a query like this now:

```gql
  getProduct(id: 2383) {
    gallery {
      image {
        srcset(thumbnail: "productImage") {
          descriptor
          url
        }
      }
    }
  }
```

...and got the expected result.

weisswurstkanone:
thx for testing!

PR #16, opened 2019-04-30
Adds the ability to return a full <img> tag

NiklasBr:
Fixes: #13 by adding a separate transformer operator that returns the full html string.


weisswurstkanone:
thx!

PR #25, opened 2019-05-09
Fixes proper Exception throwing and some minor code cleanup

NiklasBr:
* `now` -> `new` is the important part :)
* Better method docs with @params and @throws
* CamelCases clientName variable
* Removes unnecessary variable


weisswurstkanone:
thx!

Issue #15, opened 2019-04-29
Replace usage of ArrayObject with something like an "ElementDescriptor"


weisswurstkanone:

weisswurstkanone:
inherits from ArrayObject

Issue #4, opened 2019-03-20
Don't place \Pimcore::getContainer calls into the boot function of the Bundle


dpfaffenbauer:
When running tests, it might be possible that this doesn't work since the Kernel will be set later during the init of the test suite. 

Move those EventHandlers to services directly.

https://github.com/pimcore/data-hub/blob/master/src/Pimcore/Bundle/DataHubBundle/PimcoreDataHubBundle.php#L40

weisswurstkanone:
see >https://github.com/pimcore/data-hub/issues/7

PR #27, opened 2019-05-10
Feature: imageGallery implementation

### LINKED ISSUE #18: [Question] Progress on fieldcollections/galleries
We would like to continue improving this bundle, but to avoid doing duplicate work, could you tell us if any work has been done on either *Structured: Field-Collections* or *Media: Image Gallery* object data components?

NiklasBr:
This PR does not close #18, but it is a working proof of concept and I suspect it requires some changes before it is truly ready to merge. It is missing some rather important features at the moment, it cannot:

- Get Thumbnails for the images (or transformers such as Thumbnail HTML)
- Get `data` values for *Markers* and *Hotspots* (but can get the coordinates and such info

Please give it a try!

Example:

Skärmbild (140)

### DIFF
--- a/src/GraphQL/FieldConfigGenerator/Helper/ImageGallery.php
+++ b/src/GraphQL/FieldConfigGenerator/Helper/ImageGallery.php
@@ -1,0 +1,100 @@
+<?php
+
+/**
+ * Pimcore
+ *
+ * This source file is available under two different licenses:
+ * - GNU General Public License version 3 (GPLv3)
+ * - Pimcore Enterprise License (PEL)
+ * Full copyright and license information is available in
+ * LICENSE.md which is distributed with this source code.
+ *
+ *  @copyright  Copyright (c) Pimcore GmbH (http://www.pimcore.org)
+ *  @license    http://www.pimcore.org/license     GPLv3 and PEL
+ */
+
+namespace Pimcore\Bundle\DataHubBundle\GraphQL\FieldConfigGenerator\Helper;
+
+use GraphQL\Type\Definition\ResolveInfo;
+use Pimcore\Bundle\DataHubBundle\GraphQL\FieldHelper\AssetFieldHelper;
+use Pimcore\Model\Asset;
+use Pimcore\Model\DataObject\AbstractObject;
+use Pimcore\Model\DataObject\ClassDefinition;
+use Pimcore\Model\DataObject\Data\Hotspotimage;
+use Pimcore\Model\Element\AbstractElement;
+use Pimcore\Model\Element\Service;
+
+class ImageGallery
+{
+    /**
+     * @var ClassDefinition\Data\ImageGallery
+     */
+    public $fieldDefinition;
+
+    /**
+     * @var ClassDefinition
+     */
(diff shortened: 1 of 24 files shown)


src/GraphQL/FieldHelper/AbstractFieldHelper.php
```diff
@@ -115,15 +111,16 @@ public function getArguments(FieldNode $ast)
      */
     public function extractData(&$data = [], $container, $args, $context, ResolveInfo $resolveInfo = null)
     {
-        if ($container instanceof ElementInterface) {
+        if (method_exists($container, 'getId')) {
```
weisswurstkanone: why ?

NiklasBr: That might have been an artifact of a problem I had when `extractData()` got a `Hotspotimage` object which was an `ElementInterface`, I solved it by sending `$relation->getImage()` rather than `$relation` in `Helper\ImageGallery`. I should restore it I think.

src/GraphQL/FieldConfigGenerator/Helper/ImageGallery.php
```diff
@@ -0,0 +1,107 @@
...
+     * Objects constructor.
+     *
+     * @param $fieldDefinition
+     * @param ClassDefinition $class
+     */
+    public function __construct(ClassDefinition\Data $fieldDefinition, ClassDefinition $class)
+    {
+        $this->fieldDefinition = $fieldDefinition;
+        $this->class = $class;
+        $this->fieldHelper = \Pimcore::getContainer()->get('pimcore.datahub.graphql.fieldhelper.asset');
```
weisswurstkanone: we should already keep https://github.com/pimcore/data-hub/issues/4 in mind and avoid/replace it wherever easily possible

src/GraphQL/FieldConfigGenerator/Helper/ImageGallery.php
```diff
@@ -0,0 +1,107 @@
...
+
+            if ($relations) {
+                /** @var $relation AbstractElement */
+                foreach ($relations as $relation) {
+                    if ($relation instanceof Hotspotimage) {
+                        $image = $relation->getImage();
+                    }
+
+                    if ($image instanceof Asset) {
+                        $data = [];
```
weisswurstkanone: please use ArrayObject instead.
also see https://github.com/pimcore/data-hub/issues/15

reason is that we want to distinguish between ordinary arrays and something that describes an element itself using an array structure like below

NiklasBr:
I have now added support for the `data` attributes on Markers and Hotspots. I consider the most basic functionality done. Still remains to add support for thumbnails (and thumbnail html) transformers.

Skärmbild (144)

weisswurstkanone:
great, do you want to add support for the transformers as part of the existing PR or merge it first ?

NiklasBr:
Thanks! I think I prefer to merge now if possible and make those additions as a second PR.

weisswurstkanone:
thx!

Issue #43, opened 2019-06-12
Create the basis for mutations


weisswurstkanone:
* create objects
* update objects (support a few datatypes and operators)
* delete objects

* create/read/update/delete assets (at least data + filename)
* create/read/update/delete asset & object folders

NiklasBr:
If there are no Mutations configured then `GraphQL/Mutation/MutationType.php` method `build()` will throw because: *Warning: ksort() expects parameter 1 to be array, null given*

I'm making a PR.

PR #55, opened 2019-06-14
Prevents a fatal error when no mutations are configured

NiklasBr:
Also: Will not attempt to sort if there is only one mutation, see #43


weisswurstkanone:
thx!

Issue #45, opened 2019-06-13
Security/Workspace Setttings: CRUD permissions instead of RW


weisswurstkanone:

weisswurstkanone:
sorry, it is back again :-)

PR #56, opened 2019-06-14
Security/Workspace Setttings: CRUD permissions instead of RW #45

weisswurstkanone:


NiklasBr:
I believe this change undid #55 😨

PR #57, opened 2019-06-17
Avoid PHP Notice on first time a Configuration is saved

NiklasBr:
The first time a new Configuration is saved, the `$securityConfig` value can be `null`, this prevents that PHP notice.


weisswurstkanone:
thx!

PR #60, opened 2019-06-17
Renames 'childs' to 'children'

### LINKED ISSUE #59: Change Childs to Children to align with Pimcore 6
With the update to Pimcore 6 the `getChilds()` methods and `childs` attributes (among some other) have changed to the more proper `children`. Doing the same here would be nice,

I'm preparing a PR.

NiklasBr:
Resolves #59.

Also changes related verbiage while attempting to retain Pimcore 5 compatibility by checking for:
```php
$children = method_exists($this, 'getChildren') ? $this->getChildren() : $this->getChilds();
```
That should be easy enough to remove in a future update when Pimcore 5 support is dropped.

### DIFF
--- a/src/GraphQL/InputProcessor/Base.php
+++ b/src/GraphQL/InputProcessor/Base.php
@@ -72,7 +72,7 @@ class Base
      */
     public function getParentProcessor($nodeDef, ClassDefinition $class) {
         $nodeDefAttributes = $nodeDef["attributes"];
-        $children = $nodeDefAttributes["childs"];
+        $children = $nodeDefAttributes['children'];
         if (!$children) {
             return null;
         }
--- a/src/GraphQL/InputProcessor/Base.php
+++ b/src/GraphQL/InputProcessor/Base.php
@@ -72,7 +72,7 @@ class Base
      */
     public function getParentProcessor($nodeDef, ClassDefinition $class) {
         $nodeDefAttributes = $nodeDef["attributes"];
-        $children = $nodeDefAttributes['children'];
+        $children = $nodeDefAttributes['childs'];
         if (!$children) {
             return null;
         }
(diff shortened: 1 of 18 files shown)


src/GraphQL/InputProcessor/Base.php
```diff
@@ -72,7 +72,7 @@ public function process(Concrete $object, $newValue, $args, $context, ResolveInf
      */
     public function getParentProcessor($nodeDef, ClassDefinition $class) {
         $nodeDefAttributes = $nodeDef["attributes"];
-        $children = $nodeDefAttributes["childs"];
+        $children = $nodeDefAttributes['children'];
```
weisswurstkanone: I fear that this is not possible because we would also have to change it on the client side which could have a huge impact.

src/GraphQL/InputProcessor/IfEmptyOperator.php
```diff
@@ -39,27 +39,27 @@ public function __construct($nodeDef)
...
+     * @throws \Exception|\UnexpectedValueException
      */
     public function process(Concrete $object, $newValue, $args, $context, ResolveInfo $info)
     {
         $class = $object->getClass();
 
         $nodeDef = $this->nodeDef;
         $nodeDefAttributes = $nodeDef["attributes"];
-        $children = $nodeDefAttributes["childs"];
+        $children = $nodeDefAttributes['children'];
```
weisswurstkanone: same here

src/GraphQL/InputProcessor/Base.php
weisswurstkanone: see fieldConfigDialog.js (which is more or a less a copy of gridConfigDialog.js)

NiklasBr: I could have sworn I tested it (after clearing caches) and it worked, but you are correct, I am reverting these two lines

PR #67, opened 2019-06-19
Support for Calculated Value datatype

### LINKED ISSUE #66: Support Calculated values datatype
@weisswurstkanone Do you have any internal progress on this? We are considering starting on this feature because we have a use case for it.

NiklasBr:
Resolves #66.

### DIFF
--- a/src/GraphQL/QueryFieldConfigGenerator/CalculatedValue.php
+++ b/src/GraphQL/QueryFieldConfigGenerator/CalculatedValue.php
@@ -16,7 +16,6 @@
 namespace Pimcore\Bundle\DataHubBundle\GraphQL\QueryFieldConfigGenerator;
 
 use Pimcore\Model\DataObject\ClassDefinition\Data;
-use GraphQL\Type\Definition\Type;
 
 /**
  * Class CalculatedValue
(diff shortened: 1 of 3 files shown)


src/GraphQL/QueryFieldConfigGenerator/CalculatedValue.php
```diff
@@ -0,0 +1,48 @@
...
+     */
+    public function getGraphQlFieldConfig($attribute, Data $fieldDefinition, $class = null, $container = null)
+    {
+        return $this->enrichConfig(
+            $fieldDefinition,
+            $class,
+            $attribute,
+            [
+                'name' => $fieldDefinition->getName(),
+                'type' => $this->getFieldType($fieldDefinition, $class, $container),
```
weisswurstkanone: which defaults to string, right?

NiklasBr: Yes, it inherits `getFieldType()` from Base, and as far as I understand it will always be a string, [even the numerical addition example in docs returns a string ](/pimcore/pimcore/blob/master/doc/Development_Documentation/05_Objects/01_Object_Classes/01_Data_Types/33_Calculated_Value_Type.md).

weisswurstkanone: True, which is kind of sad because the only reason for that is that it needs a string for displaying it in the grid + for csv export. Maybe this restriction should be relaxed.

weisswurstkanone:
thx a lot!

PR #69, opened 2019-06-21
Add support for Input Quantity Value query

dvesh3:


weisswurstkanone:
doesn't work if there is more than one input qv field

PR #74, opened 2019-06-26
Updates webserviceController to strengthen security check

NiklasBr:
and throw more explicit error. performSecurityCheck() now returns void because the return value is never used anyway


weisswurstkanone:
thx!

PR #86, opened 2019-07-09
Bugfix: Hotspotimage renamed broke imagegalleries

NiklasBr:
Fix for **unknown type: hotspot** thrown in [GraphQL/Service.php](https://github.com/pimcore/data-hub/blob/master/src/GraphQL/Service.php#L449), introduced due to change in graphql.yml


weisswurstkanone:
thx!

PR #92, opened 2019-07-16
Updates QueryType to align with array arguments

NiklasBr:
Even if the GQL would complain about using a string where an array of strings were used in the query the listing would accept it, this caused a slight mismatch when running automated queries.

### DIFF
--- a/src/GraphQL/Resolver/QueryType.php
+++ b/src/GraphQL/Resolver/QueryType.php
@@ -292,10 +292,11 @@ class QueryType
         }
 
         // sorting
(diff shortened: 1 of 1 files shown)


weisswurstkanone:
explain "GQL would complain" :-)

NiklasBr:
Skärmbild (162)

But actually sending this query (e.g. using Curl) would work (when the schema says it should not). It's just a UI thing where the query should properly align with the the schema.

weisswurstkanone:
Exactly. I would go for it. Any other opinions ?

PR #98, opened 2019-08-20
Removes possibly errounous negation in $config

NiklasBr:
I was looking for something else and found this, I just assume it was an unnoticed bug


weisswurstkanone:
LOL. many thanks! :-)

PR #99, opened 2019-08-21
Config improvements

### LINKED ISSUE #97: Configuration verbosity problems
Every time the `datahub-configurations.php` file is saved a lot of (what seems to be) unnecessary fields change, for example

```
"key" => "#5d5a91383afe1"
```
and
```
"displayfield-2452-inputEl" => "API key required for web service access (min length: 16)"
```

Are these required for the config to work or can the file be generated without them?

NiklasBr:
Resolves #97 

However, this PR does not yet touch the `"key" => "#5d..."` in datahub-configuration.php (because I don't understand it yet) but does a number of other things:

* Adds return types, implements class constant for reused string, normalises Exception class import according to comments, moves to shorthand ifs (`?:`) for configController.php
* Adds return types, normalises Exception throwing, adds access keywords, and method docs for Dao.php
* Adds return types, simplifies IsInstalled() logic, uses ConfigController constant for name in Installer.php
* Prevents SQL hint and API info from saving in datahub-configuration.php

### DIFF
--- a/src/Controller/ConfigController.php
+++ b/src/Controller/ConfigController.php
@@ -19,7 +19,6 @@ use Pimcore\Bundle\DataHubBundle\Configuration;
 use Pimcore\Bundle\DataHubBundle\Configuration\Dao;
 use Pimcore\Bundle\DataHubBundle\GraphQL\Service;
 use Pimcore\Bundle\DataHubBundle\WorkspaceHelper;
-use Symfony\Component\Config\Definition\Exception\Exception;
 use Symfony\Component\HttpFoundation\JsonResponse;
 use Symfony\Component\HttpFoundation\Request;
 use Symfony\Component\Routing\Annotation\Route;
@@ -30,7 +29,16 @@ use Symfony\Component\Routing\RouterInterface;
  */
 class ConfigController extends \Pimcore\Bundle\AdminBundle\Controller\AdminController
 {
-    private function buildFolder($path, $name)
+    public const CONFIG_NAME = 'plugin_datahub_config';
+
+    /**
+     * @param $path
+     *
+     * @param $name
+     *
+     * @return array
+     */
+    private function buildFolder($path, $name): array
     {
         return [
             'id' => $path,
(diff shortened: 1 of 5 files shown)


NiklasBr:
Before and after:
Skärmbild (169)

src/Controller/ConfigController.php
```diff
@@ -124,25 +128,27 @@ public function listAction(Request $request)
...
-        $this->checkPermission('plugin_datahub_config');
+        $this->checkPermission(self::CONFIG_NAME);
 
         try {
             $name = $request->get('name');
 
             $config = Dao::getByName($name);
-            if (empty($config)) {
-                throw new Exception('Name does not exist.');
+            if ($config instanceof Configuration) {
```
weisswurstkanone: really ?

NiklasBr: `getByName()` is documented as returning `@return Configuration|null` (I also checked manually) and I followed PHPStorm's inspection recommendation:

"'empty(...)' counts too many values as empty, consider refactoring with type sensitive checks."

Since `getByName()` should only return a matching object or null and null will never be an instance of anything the code works.

weisswurstkanone: I am referring to this change

            if ($config instanceof Configuration) {
                throw new \Exception('Name does not exist.');
            }

NiklasBr: I was unsure whether or not you wanted to use `\Exception` or `Symfony\Component\Config\Definition\Exception\Exception` but settled for the former because `Pimcore\Bundle\DataHubBundle\Configuration` throws `\Exception` and the docs in this file said `@throws \Exception` [here](https://github.com/pimcore/data-hub/blob/master/src/Controller/ConfigController.php#L244) and [here](https://github.com/pimcore/data-hub/blob/master/src/Controller/ConfigController.php#L277) and [here](https://github.com/pimcore/data-hub/blob/master/src/Controller/ConfigController.php#L367).

That is why I normalised the usage to \Exception for the Dao-related code.

weisswurstkanone: Still, maybe the ! is missing. If $config is instanceof Configuration then it exists. The exception claims it does not exist.

NiklasBr: Oh, right! 😄

weisswurstkanone:
ready to be merged ?

NiklasBr:
Yup!

weisswurstkanone:
thanks a lot!

Issue #111, opened 2019-09-30
API-Platform instead of directly using GraphQL


dpfaffenbauer:
I thought a lot about integrating API-Platform instead of the graphql library in the last couple of days.

IMO: I don't see any reason why it shouldn't work, and, it would also solve the REST feature automatically.

Things that need to be done in order for API-Platform to work with DataHub:

 - Map the DataHub Class Configuration to API-Platform Metadata
 - Add an EntryPoint for API-Platform to load DataHub Config
 - Create DataObject CollectionProvider to load listings based on MetaData (respect workspace Configuration, Fields and Transformers)
 - Create DataObject ItemProvider (respect workspace Configuration, Fields and Transformers)
 - Create DataObject DataPersister (respect workspace Configuration, Fields and Transformers)
 - Create DataObject Filters
 - Create CollectionProvider, ItemProvider and DataPersister for "simple" entities (Documents, Assets)

I know, we had this discussion before, but I don't understand the reason why it wasn't implemented with API-Platform in the first place...

dpfaffenbauer:
Proof of Concept:
<img width="1241" alt="Bildschirmfoto 2019-09-30 um 18 08 29" src="https://user-images.githubusercontent.com/5981845/65896410-76978680-e3ad-11e9-986a-9afde369e5eb.png">
![Bildschirmfoto 2019-09-30 um 18 08 56](https://user-images.githubusercontent.com/5981845/65896391-6c758800-e3ad-11e9-9aaf-e8aeebcb641d.png)

Config
![Bildschirmfoto 2019-09-30 um 18 10 02](https://user-images.githubusercontent.com/5981845/65896445-8d3ddd80-e3ad-11e9-9212-5a680dd71aec.png)

I will create a WIP PR for you to checkout.

Cruiser13:
This is awesome. We're currently working on a more direct integration of api platform with pimcore.
We took a different path (generating custom classes which do extend the pimcore dataobject classes and have the annotations for api platform included) which will then be extended with custom controller classes which handle stuff like permissions and everything.

Your approach is far better with this direct data hub integration.

Cruiser13:
Using API-Platform would also solve #93 and #94 as API Platform does support them natively:
https://api-platform.com/docs/core/content-negotiation/

dpfaffenbauer:
[...]

PR #112, opened 2019-09-30
WIP: API-Platform Proof Of Concept

dpfaffenbauer:
Discussion: https://github.com/pimcore/data-hub/issues/111

This is JUST a proof of concept. It is possible to connect Pimcore DataHub with API-Platform and get the best of booth worlds: REST and GraphQL, and, also have the possibility of a custom configuration UI, which I think is great.

This works just very basic, operators and stuff are not supported, but it should just proof the possibility of using it as a base framework for this.

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -13,7 +13,8 @@
   },
   "require": {
     "pimcore/pimcore": "^5.7 || ^6.0",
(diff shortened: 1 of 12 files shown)


dpfaffenbauer:
About the CLA thing: it seems to be broken for now... I'll check back later

PR #117, opened 2019-10-15
Add EventListener to dedicated EventListener Class

solverat:
Beside of improving symfony standards, this will prevent exceptions like these (while executing tests for example):

image

### DIFF
--- a/src/Command/GraphQL/RebuildDefinitionsCommand.php
+++ b/src/Command/GraphQL/RebuildDefinitionsCommand.php
@@ -14,7 +14,6 @@
 
 namespace Pimcore\Bundle\DataHubBundle\Command\GraphQL;
 
-use Elements\Bundle\ProcessManagerBundle\ExecutionTrait;
 use Pimcore\Bundle\DataHubBundle\Configuration;
 use Pimcore\Console\AbstractCommand;
 use Symfony\Component\Console\Input\InputInterface;
@@ -23,10 +22,8 @@ use Symfony\Component\Console\Output\OutputInterface;
 
 class RebuildDefinitionsCommand extends AbstractCommand
 {
-
-
     /**
-     *
+     * {@inheritDoc}
      */
     protected function configure()
     {
@@ -38,8 +35,7 @@ class RebuildDefinitionsCommand extends AbstractCommand
                 null,
                 InputOption::VALUE_REQUIRED,
             'Comma separated list of endpoints'
-            )
-        ;
+            );
     }
 
     /**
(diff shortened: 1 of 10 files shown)


weisswurstkanone:
Hi, two things
- please don't forget the documents
- second and more important: the master configuration is still saved in datahub-configurations.php
Workspace permissions are written to database just for performance reasons.

solverat:
@weisswurstkanone:

1. Alright, I just added them (I didn't really forget them, they just wasn't there before).
2. I don't really know what to do with this fact: Is there anyhing I need to do?

solverat:
@weisswurstkanone ah! you're talking about those elements, right:

image

Shouldn't we use IDs instead of paths in the first place? Could save us some troubles (like the current rename issue for example)

solverat:
Second thought: IDs would be nice but it wouldn't solve the problem. 

I'll add a method which will check the configuration file like you suggested.

solverat:
@weisswurstkanone: Alright, the latest commit:

- checks configuration file for changed values (modified or deleted). If so, adopt the changes
- checks modification date of configuration file (similar to the class definition editor). if there are mismatches, force user to reload config

NiklasBr:
> Shouldn't we use IDs instead of paths in the first place? Could save us some troubles (like the current rename issue for example)

That saves some trouble, absolutely. But it also adds potential problems elsewhere: Running test environments where almost daily builds and reinstallations cause the database id:s of objects to change.

solverat:
@NiklasBr: i know, That‘s why i kept it that way.

dpfaffenbauer:
We had that ID/Path discussion already many times. Since an ID is environment specific, we always have to work with Paths instead. Thats basically the only reason

weisswurstkanone:
As I said I think that manipulating the database directly is not necessary as this is done implicitely by Configuration::save

see https://github.com/pimcore/data-hub/blob/48047edc056d946b80c51de020d0a54146738926/src/Configuration.php#L173

The existing event handler is probably just a leftover.

solverat:
@weisswurstkanone : that's not entirely true. If you're changing or deleting a document/asset/object directly, you also need to update the db workspaces (the `saveWorkspaces` only updates the db if you modify the configuration itself.) You can check that by disabling the event listener.

weisswurstkanone:
Yes that's absolutely true and I am aware of that. I was just saying that we shouldn't fix do things twice since we need the existing db update logic when you save an existing configuration :-)

weisswurstkanone:
thanks a lot!

PR #119, opened 2019-10-16
Replace plugin to bundle in frontend namespace

vmalyk:
1. Changed old namespace from Pimcore 4 (.plugin) to Pimcore 5/6 (.bundle) for consistency
2. Removed extra lines in js files

### DIFF
--- a/doc/graphl/AddCustomMutationOperator.md
+++ b/doc/graphl/AddCustomMutationOperator.md
@@ -29,7 +29,7 @@ and the server-side PHP implementation processing the input (the input processor
 A JS sample can be found here:
 https://github.com/pimcore/data-hub/blob/master/src/Resources/public/js/mutationoperator/IfEmpty.js
 
-Note that the namespace in your case would be `pimcore.plugin.datahub.mutationoperator.mycustommutationoperator`
+Note that the namespace in your case would be `pimcore.bundle.datahub.mutationoperator.mycustommutationoperator`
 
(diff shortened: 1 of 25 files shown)


dpfaffenbauer:
not sure how Pimcore manages SemVer with this Bundle, but this is considered a BC break. CoreShop already implements it's own data-types into data-hub. With this PR, it will stop working

dpfaffenbauer:
I mean, it's not a full 1.0 yet... so BC breaks are allowed, but still....

weisswurstkanone:
With this change the datahub UI vanishes completely

weisswurstkanone:
Probably because of this line

pimcore.bundle.datahub = Class.create(pimcore.bundle.admin, {

weisswurstkanone:
see https://pimcore.com/docs/5.x/Development_Documentation/Extending_Pimcore/Bundle_Developers_Guide/Plugin_Backend_UI.html

weisswurstkanone:
Hi, now configurations cannot be opened.
Did you do any hands-on testing or just search & replace ?

Regardless of that, is there anybody out there with objections against the "BC break"? Or in other words, are there any custom operator/data type implementations already?

dpfaffenbauer:
Yes, CoreShop ;)

vmalyk:
@weisswurstkanone and @dpfaffenbauer, if you think these changes are not finalized, BC or irrelevant - I declined PR. By the way, CoreShop implemented some classes on Early preview module? If it`s true, some some another changes `ll be break  theirs  code ( and they know that(and not only CoreShop)) on non released version (ex 1.0.0).

dpfaffenbauer:
Not sure what you mean....

PR #122, opened 2019-10-24
Fix Assets fullpath(thumbnail:) and data(thumbnail:) 

### LINKED ISSUE #121: Assets fullpath(thumbnail:) not working on relations
I'm trying to get thumbnail from asset linked by relation to an object, but I cannot get it to work. I tried to check over the code but without results.

To reproduce, this is a simple class:
image

This is an object created with that class:
image

This is the schema in datahub config:
image

This is a sample GraphQl Query tho show the error:
image

The field "thumb" in the asset linked by the relation is always the original asset path, never the thumbnail path. (As you can see in the direct getAsset query).

SimoneCarnio:
Should fix #121

### DIFF
--- a/src/GraphQL/AssetType/AssetType.php
+++ b/src/GraphQL/AssetType/AssetType.php
@@ -91,6 +91,14 @@ class AssetType extends ObjectType
                                 return $image->getFullPath();
                             }
                         }
+                        if ($image instanceof Asset\Document)
+                        {
+                            if (isset($args["thumbnail"])) {
+                                return $image->getImageThumbnail($args["thumbnail"]);
+                            } else {
+                                return $image->getFullPath();
+                            }
+                        }
                     }
                 }
             ],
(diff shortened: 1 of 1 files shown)


src/GraphQL/AssetType/AssetType.php
```diff
@@ -104,7 +112,39 @@ public function build(&$config)
...
+                                $thumb = $image->getImageThumbnail($args['thumbnail']);
+                                return base64_encode(file_get_contents($thumb->getFileSystemPath()));
+                            } else {
+                                return base64_encode(file_get_contents($image->getFileSystemPath()));
+                            }
+                        }
+                    }
+                    return null;
+                }
+            ],
```
weisswurstkanone: I think there is something wrong with the brackets here. Could you please double-check ?

weisswurstkanone:
thx!

PR #124, opened 2019-10-25
Correctly assign locale service when building data object mutations

### LINKED ISSUE #123: Mutation not working as soon as defaultLanguage used
I have a mutation which can be executed successfully as long as I don't specify the defaultLanguage.

When the defaultLanguage is specified, an error is thrown here:
https://github.com/pimcore/data-hub/blob/master/src/GraphQL/Mutation/MutationType.php#L389
because `$localeService` is null.

Any help would be kindly appreciated!

Ben305:
Fixes #123


weisswurstkanone:
Thanks a lot for your investigations!

PR #148, opened 2019-12-09
Allow CORS requests

julkue:
This should help accessing DataHub from e.g. localhost on DEV, but also on PROD when the frontend is hosted under e.g. XXX.com and the backend on YYY.com

### DIFF
--- a/src/Controller/WebserviceController.php
+++ b/src/Controller/WebserviceController.php
@@ -170,6 +170,11 @@ class WebserviceController extends FrontendController
             ];
         }
 
+        header('Access-Control-Allow-Origin: ' . $_SERVER['HTTP_ORIGIN']);
+        header('Access-Control-Allow-Credentials: true');
+        header('Access-Control-Allow-Methods: GET, POST, PATCH, PUT, DELETE, OPTIONS');
+        header('Access-Control-Allow-Headers: Origin, Content-Type, X-Auth-Token');
+        
         return new JsonResponse($output);
     }
 
--- a/src/Controller/WebserviceController.php
+++ b/src/Controller/WebserviceController.php
@@ -169,9 +169,12 @@ class WebserviceController extends FrontendController
                 ],
             ];
         }
-
-        header('Access-Control-Allow-Origin: ' . $_SERVER['HTTP_ORIGIN']);
-        header('Access-Control-Allow-Credentials: true');
+        
+        $origin = '*';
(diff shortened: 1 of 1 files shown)


src/Controller/WebserviceController.php
```diff
@@ -170,6 +170,11 @@ public function webonyxAction(Service $service, LocaleServiceInterface $localeSe
             ];
         }
 
+        header('Access-Control-Allow-Origin: ' . $_SERVER['HTTP_ORIGIN']);
```
NiklasBr: Are you sure `HTTP_ORIGIN` is always available? See 
https://stackoverflow.com/questions/41231116/serverhttp-origin-doesnt-work
Consider maybe use `Access-Control-Allow-Origin: *`?

julmot: Thanks for your input, I agree it would make sense to have a fallback. The reason why I've decided for HTTP_ORIGIN over * is because * won't work when credentials are sent. For example, on our staging platform, we're securing it with basic auth. To avoid entering the credentials twice, we're using the credentials forwarding option. But this won't work with *.

> For requests without credentials, the literal value "*" can be specified, as a wildcard; the value tells browsers to allow requesting code from any origin to access the resource. Attempting to use the wildcard with credentials will result in an error.

See https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Access-Control-Allow-Origin

For the case when it's undefined, we can fall back to *, what do you think?

NiklasBr:
Are you sure `HTTP_ORIGIN` is always available? See 
https://stackoverflow.com/questions/41231116/serverhttp-origin-doesnt-work
Consider maybe use `Access-Control-Allow-Origin: *`?

julmot:
Not sure why my comment was deleted, I have answered your question.

weisswurstkanone:
@NiklasBr , still any concerns ?

NiklasBr:
I have no concerns other than maybe limit the allowed methods to `POST, GET, OPTIONS` the others are never used as far as I know.

weisswurstkanone:
@julmot , what do you think ?

julmot:
I've dropped PATCH, PUT and DELETE.

weisswurstkanone:
thx a lot

ishaan1994:
Hi Team, 
Has this code change been released in any data-hub version? 
Because, setting headers this way using php header function is working fine, but in the latest version i.e 1.5.1 headers are being send in the response object which is still giving the CORS Error.

PR #150, opened 2019-12-16
Parent children sibling objects

juckerf:
This PR makes object hierarchies (e.g. parents, children, siblings) available in GraphQL queries.
An example query looks like this:
```
          query {
            product: getProduct(id: 21) {
              ... on object_Product {
                parent {
                  ... on object_Product {
                    id
                    name
                    children {
                      ... on object_Product {
                        id
                        name
                      }
                    }
                  }
                }
                siblings {
                  ... on object_Product {
                    id
                    name
                  }
                }
                id
                name
              }
            }
          }
```


weisswurstkanone:
Thx! Apart from the general idea (same for documents, ... would be missing I guess), does it consider workspace permissions?
What do others think?

juckerf:
@weisswurstkanone You're right, documents or even folders are not covered in this PR yet nor does it cover any mutations. It also allows exploitation of workspace permissions which are narrowed down to a specific path by traversing up the three (something I definitely have to fix).

juckerf:
We're now able to get the parent, children and siblings on any common element type (assets, documents, objects) and workspace permissions are checked.
(@mugge6 still has to sign the CLA)

src/GraphQL/AssetType/AssetFolderType.php
```diff
@@ -64,7 +64,19 @@ public function build(&$config)
...
+            ],
+            'parent' => [
+                'type' => $this,
+                'resolve' => [$elementResolver, "resolveParent"],
+            ],
+            'children' => [
+                'type' => Type::listOf($this),
+                'resolve' => [$elementResolver, "resolveChildren"],
+            ],
+            'siblings' => [
```
weisswurstkanone: siblings is not a reserved word so it can clash with a pimcore class attribute

weisswurstkanone:
Could please add the sample from above to the documentation ?

src/GraphQL/DataObjectType/PimcoreObjectType.php
```diff
@@ -92,11 +92,35 @@ public function build($context = [])
...
                         'type' => Type::listOf(Type::string()),
-                        'description' => 'comma seperated list of key names'
+                        'description' => 'comma separated list of key names'
                     ]
                 ],
                 'resolve' => [$resolver, "resolveProperties"]
-            ]
+            ],
+            'parent' => [
+                'type' => $this,
```
weisswurstkanone: Isn't this way to restrictive ? Parent, sibling & children can be of any type (at least the ones enabled in the schema), right ?

weisswurstkanone: Funny thing is that this is possible (see demo.pimcore.fun) although the parent is a folder.
It returns the fullpath (I added it to the Car schema) but bombs out if I try to add a Car attribute :-)

{
  getCar(id: 260) {
  	name
    parent {      
      fullpath
    }    
  }
}

As said, parent should be a union type (allowed classes + folder) and children/siblings should be a list of that type.

src/GraphQL/Resolver/Element.php
```diff
@@ -79,5 +72,151 @@ public function resolveProperties($value = null, $args = [], $context, ResolveIn
...
+     * @param array $args
+     * @param $context
+     * @param ResolveInfo|null $resolveInfo
+     * @return array
+     * @throws \Exception
+     */
+    protected function extractSingleElement($element, $args, $context, $resolveInfo)
+    {
+        if (!WorkspaceHelper::isAllowed($element, $context['configuration'], 'read')) {
+            if (PimcoreDataHubBundle::getNotAllowedPolicy() == PimcoreDataHubBundle::NOT_ALLOWED_POLICY_EXCEPTION) {
```
weisswurstkanone: "unknown class PimcoreDataHubBundle" ... import missing ?

src/GraphQL/AssetType/AssetFolderType.php
juckerf: what do you suggest? adding sibling(s) to the reserved words list (which would be a breaking change and would have to be done in [the core](https://github.com/pimcore/pimcore/)) or use another key (seems like none of the reserved words fits for this) or discard the functionality of getting siblings via data-hub for now?

src/GraphQL/Resolver/Element.php
juckerf: my bad. will fix it asap

src/GraphQL/DataObjectType/PimcoreObjectType.php
juckerf: thanks for the feedback. I think you're right with using a union type (this restriction goes back when we designed the feature only with objects in mind).

src/GraphQL/AssetType/AssetFolderType.php
weisswurstkanone: I think you could prepend it with an underscore (which would also make it clear that it is not part of the data model :-)

src/GraphQL/DataObjectType/PimcoreObjectType.php
juckerf: @AlsonPL almost resolved the union types issue (only an union type for documents/folders etc. is missing). We'll deliver this asap

AlsonPL: We're now able to get the parent, children and siblings on asset, object and documents, and also on asset folders, object folders and document folders.

weisswurstkanone: meaning that it would be ready to me merged assumed that everything is fine ? 😃

AlsonPL: I just added sample query to documentation. I hope everything is ok and it's ready to be merged :)

src/GraphQL/DataObjectType/ObjectTreeType.php
```diff
@@ -0,0 +1,57 @@
...
+        parent::__construct($config);
+    }
+
+    /**
+     * @return array
+     * @throws \Exception
+     */
+    public function getTypes()
+    {
+        $types = array_values(ClassTypeDefinitions::getAll(false));
```
weisswurstkanone: shouldn't this be getAll(true) instead ?

src/GraphQL/DataObjectType/ObjectTreeType.php
```diff
@@ -0,0 +1,57 @@
...
+    }
+
+    public function resolveType($element, $context, ResolveInfo $info)
+    {
+        if (!$element) {
+            return null;
+        }
+        $object = DataObject\Concrete::getById($element['id']);
+
+        if ($object instanceof DataObject\Folder) {
```
weisswurstkanone: I guess Concrete::getById() can never DataObject\Folder

doc/graphl/querysamples/Sample_ParentChildrenSiblings.md
```diff
@@ -0,0 +1,79 @@
...
+    name
+    
+    parent {
+      ... on object_Car {
+        id
+        name
+      }
+    }
+    
+    children {
```
weisswurstkanone: please mention the possibility to filter the type (folder, variant, ...) as there was already a question about that

src/GraphQL/Resolver/Element.php
```diff
@@ -81,5 +74,152 @@ public function resolveProperties($value = null, $args = [], $context, ResolveIn
...
+        return $arguments;
+    }
+
+    /**
+     * @param $value
+     * @return AbstractElement|null
+     * @throws \Exception
+     */
+    protected function getElementById($id)
+    {
```
weisswurstkanone: we probably can replace this entire block with Pimcore\Model\Element\Service::getElementById()

weisswurstkanone:
@juckerf ping :-)

juckerf:
@weisswurstkanone thanks for the review. I was quite busy the past days. will check your comments asap

weisswurstkanone:
great, thank you!

romainbessugesmeusy:
Hi @weisswurstkanone, I tested the changes locally and it seems to work as expected. Do you have some time to take a look at the PR? Thank you very much

src/GraphQL/Resolver/Element.php
```diff
@@ -81,5 +75,132 @@ public function resolveProperties($value = null, $args = [], $context, ResolveIn
...
+            $parent = $element->getParent();
+            if ($parent) {
+                return $this->extractSingleElement($parent, $args, $context, $resolveInfo);
+            }
+        }
+        return null;
+    }
+
+    /**
+     * @param null $value
```
weisswurstkanone: array

romainbessugesmeusy: Sorry I don’t understand what you mean

weisswurstkanone: I believe the phpdoc type hint isn't right. It should be array instead of null.

weisswurstkanone:
Following query fails. Seems that this is always the case if there is an element in the children list which is not allowed by the schema definition.

image

weisswurstkanone:
image

weisswurstkanone:
Works fine for children of kind "Car" for example

src/GraphQL/Resolver/Element.php
juckerf: I fixed the hint (btw: `$value` has a default value of null which is not checked in the added methods, but same applies for the existing method `resolveProperties`)

juckerf:
> Following query fails. Seems that this is always the case if there is an element in the children list which is not allowed by the schema definition.
> 
> image

I guess this is because of the `true` parameter in https://github.com/pimcore/data-hub/blob/c54be80bc816cc00dc854bf17588fba3159e7c41/src/GraphQL/DataObjectType/ObjectTreeType.php#L38 which only allows types/entities which are configured in the schema

weisswurstkanone:
Guess then we should filter them out if not part of the schema.

romainbessugesmeusy:
Hallo @juckerf 
Do you know if the requested changes are in the making? This feature would be a game changer for our PIM. Maybe we can help if you lack the workforce?

juckerf:
@romainbessugesmeusy 
I'm planing to complete the requested changes in the next few days

juckerf:
@weisswurstkanone elements which are not exposed in the schema are now excluded in the parent/sibling/children resolver

weisswurstkanone:
Thanks, found one thing:

image

weisswurstkanone:
Maybe double-check the yml.

      tags:
          - { name: pimcore.datahub.graphql.assettype_factory, id: asset_tree }


versus
generaltype_factory

juckerf:
the asset_tree type is now registered in the proper factory

weisswurstkanone:
Thx!

This query returns 5 child assets (compare to demo) although Asset support isn't ticked in the schema (only folders).

{
  getAssetFolder(id: 335) {
    fullpath
    children {
      __typename
      ... on asset {
        creationDate
        id
      }
    }
  }
}

juckerf:
Seems like a bug in WorkspaceHelper::isAllowed which allows access to an element regardless of its type if access to a parent of the element is allowed

weisswurstkanone:
Not sure if I can follow you. I guess the asset shouldn't be even allowed to be there as is it is disabled in the schema definition. We had a similar discusssion with objects already. 

image

So this setting decides whether you can work with assets in general.
The workspace permission then decides which kind of access is allowed (on element level).

image

juckerf:
yep, you're right. I mixed up the two concepts (workspace permissions and special entity schema settings). But now it should be fixed aswell

weisswurstkanone:
Brilliant, thanks a lot!

PR #167, opened 2020-03-03
fixed faulty import

tiemenv:
When creating a GraphQL API for some object classes in Datahub, we get HTTP 500 responses when we open up fields from an entity in our query schema. Internally, PHP gives 

```PHP Fatal error:  Interface 'Pimcore\Bundle\DataHubBundle\GraphQL\DataObjectQueryFieldConfigGeneratorInterface' not found in /var/www/tiemen.pimcore6demo.studioemma.com/versions/wwwroot/vendor/pimcore/data-hub/src/GraphQL/DataObjectQueryFieldConfigGenerator/Base.php on line 26```

Exploring Base.php, I found it to be importing 

```use Pimcore\Bundle\DataHubBundle\GraphQL\DataObjectQueryFieldConfigGeneratorInterface;```

Whereas the file is called ```QueryFieldConfigGeneratorInterface.php```

Fixing this import to ```use Pimcore\Bundle\DataHubBundle\GraphQL\QueryFieldConfigGeneratorInterface;``` has fixed the issue.

### DIFF
--- a/src/GraphQL/DataObjectQueryFieldConfigGenerator/Base.php
+++ b/src/GraphQL/DataObjectQueryFieldConfigGenerator/Base.php
@@ -16,7 +16,7 @@
 namespace Pimcore\Bundle\DataHubBundle\GraphQL\DataObjectQueryFieldConfigGenerator;
 
 use GraphQL\Type\Definition\Type;
-use Pimcore\Bundle\DataHubBundle\GraphQL\DataObjectQueryFieldConfigGeneratorInterface;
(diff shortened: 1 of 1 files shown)


weisswurstkanone:
seems that you are not on the latest release. 

already fixed via https://github.com/pimcore/data-hub/commit/c41a1f0e294d1ae972a335c2fba00d51945fc266

weisswurstkanone:
however, the file is still not named correctly

weisswurstkanone:
please check if it works with the latest commit instead

tiemenv:
Everything works after the rename

weisswurstkanone:
thx for checking

PR #173, opened 2020-03-13
[Events] Pass info from pre-build events

### LINKED ISSUE #172: Custom queries
I was searching for a way to implement a custom GraphQL query which have some extra arguments and a custom resolver, but this seems not possible.

There is some event dispatched `QueryEvents::PRE_BUILD` before all the build in queries are registered, however I cannot alter the `$config` there as it is not passed by reference to the event object.

Also in the `WebserviceController` a new instance of `QueryType` is created without using any DI, which is really bad design practice. Better would be to have some factory object, ie `QueryTypeFactory` which is passed as a dependency to the webservice controler and is responsible for the creation of the query type.

Do you have a solution for my use case?

dvesh3:
resolves #172

### DIFF
--- a/src/GraphQL/Mutation/MutationType.php
+++ b/src/GraphQL/Mutation/MutationType.php
@@ -91,6 +91,7 @@ class MutationType extends ObjectType
      */
     public function build(&$config = [], $context = [])
     {
+        $config["fields"] = [];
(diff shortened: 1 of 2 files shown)


weisswurstkanone:
thanks @dvesh3

PR #176, opened 2020-03-18
Fix fatal error when using imagegallery inside fieldcollection

bramstroker:
When a ImageGallery or Hotspotimage field is added to a fieldcollection and we are trying to retrieve it with the GraphQL API, the following error occured:

`Argument 4 passed to Pimcore\Bundle\DataHubBundle\GraphQL\DataObjectQueryFieldConfigGenerator\Helper\ImageGallery::__construct() must be an instance of Pimcore\Model\DataObject\ClassDefinition, instance of Pimcore\Model\DataObject\Fieldcollection\Definition given, called in /Users/bram/www/pim-dekbeddiscounter/vendor/pimcore/data-hub/src/GraphQL/DataObjectQueryFieldConfigGenerator/ImageGallery.php on line 71`

Unfortunately `DataObject\Fieldcollection\Definition` and `DataObject\ClassDefinition` do not have a common interface which can be type hinted. So the only resolution right now is to remove the typehint.
On all the other helpers also no type hints are defined so my guess is you ran into the same issue.


weisswurstkanone:
Guess it can be still added to the phpdoc using the or operator.

     * @param ClassDefinition|Fieldcollection/Definition $class

bramstroker:
Yes we can.
The fieldcollection typehint could also be readded, no need to remove that one.
Also cleaned up the docblocks a bit.

weisswurstkanone:
please don't forget to sign the CLA. thx!

bramstroker:
I have signed the CLA yesterday. Can this be merged?

weisswurstkanone:
thanks a lot!

PR #181, opened 2020-03-26
Implement table and structuredTable query types

bramstroker:


weisswurstkanone:
Sounds cool. Could you please update the documentation as well, just mention it in the list 👍  Thx!

https://github.com/pimcore/data-hub/blob/master/doc/graphl/DataObjectQueries.md

bramstroker:
Done

src/GraphQL/DataObjectQueryFieldConfigGenerator/AbstractTable.php
```diff
@@ -0,0 +1,102 @@
...
+        }
+
+        $columnKeys = $this->getTableColumnKeys($fieldDefinition);
+        if (empty($columnKeys)) {
+            return Type::string();
+        }
+
+        $fields = [];
+        foreach ($columnKeys as $key) {
+            $fields[$key] = Type::string();
```
weisswurstkanone: as you can define a type for structured table column (Number, Text, Checkbox) can we do this here as well?

bramstroker: Yes I think so. Will have a look into this wednesday

bramstroker:
Hi, I have updated the PR to support 3 different column types for the structuredTable.
First I had used `Type:int()` for the number column, but while testing I was able to enter for example `5.5` manually in the Pimcore interface in this column, so I changed this to GraphQL type `Type:float()`.

weisswurstkanone:
Thanks a lot!

PR #182, opened 2020-03-27
Add title and description fields on documents

DzemalBecirevic:
Add title and description fields on documents


weisswurstkanone:
I believe only a page can have a title and description (but no Link for example).

DzemalBecirevic:
> I believe only a page can have a title and description (but no Link for example).

I've moved the title and description to the PageSnippetType instead of AbstractDocumentType.

weisswurstkanone:
thx!

PR #193, opened 2020-04-08
Adding new missing fields to LinkDataType.

DzemalBecirevic:
Target is named windowTarget because Target is already in use.


weisswurstkanone:
thanks @DzemalBecirevic !

PR #195, opened 2020-04-08
Use symfony requests and response objects

bramstroker:
While writing some functional tests for my client using Codeception I stumbled upon some behaviour in the `WebserviceController` which makes it almost impossible to write a functional test, without al kind of strange hacks, if even possible.

My setup is the following:
- enabled the `Pimcore\Tests\Helper\Pimcore` codeception module, which can act as an `InnerBrowser` and directly do requests to the Symfony kernel
- Wrote a `GraphQL` module which uses the `Pimcore` module to dispatch GraphQL requests.

The implementation described above expects the Symfony response and request objects to be used. This is good practice anyway as you are not expected to directly access superglobals or headers in a Symfony application as this could cause unwanted side effects.

I have refactored `php:://input` stream to use the request content now. I don't see why we need the raw input stream in this case, but maybe there is a good reason I don't know of.
And headers are set using the response object.

I have also tested in Insomnia (GraphQL client) and still get good responses and headers are identical.

### DIFF
--- a/src/Controller/WebserviceController.php
+++ b/src/Controller/WebserviceController.php
@@ -89,7 +89,6 @@ class WebserviceController extends FrontendController
         $queryType = new QueryType($service, $localeService, $modelFactory, $this->eventDispatcher, [], $context);
         $mutationType = new MutationType($service, $localeService, $modelFactory, $this->eventDispatcher, [], $context);
 
-
         try {
             $schemaConfig = [
                 'query' => $queryType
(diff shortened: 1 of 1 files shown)


weisswurstkanone:
This controller started basically with aq copy of webonyx's endpoint implementation, that's the only reason 😄

weisswurstkanone:
thx for refactoring!

PR #196, opened 2020-04-10
Added step to verify installation

florian25686:
I had issues during installation and think this aids for helping others with the same problem


weisswurstkanone:
thx @florian25686 !

Issue #203, opened 2020-04-22
"Internal Server Error" using Mutation "create" and mandatory translated field


ttheuerkauf:
I have a problem when trying to create a new object by mutation "create", if this object has a mandatory field which is translated (into two languages):

![grafik](https://user-images.githubusercontent.com/48792043/79962912-f8a1e680-8488-11ea-8248-fcdb46967018.png)

If I switch the field "name" to non mandatory, the object can be created correctly.

This behaviour is in my opinion somewhat clear, because there is no way to input the field name in all languages into the create-mutation. The value for the non-default language is simply missing, therefore Pimcore throws an exception which then results in the Internal Server Error.

Is there any undocumented way to perform this mutation correctly?

weisswurstkanone:
Currently no. Maybe we should add a flag to the create operation.
Can you create a PR please ?

Dirty workaround for now could be:
- add a listener to ExecutorEvents::PRE_EXECUTE
- in the event handler add a listener to DataObjectEvents::PRE_ADD and call $e->getObject()->setOmitMandaroryCheck(true)

PR #204, opened 2020-04-22
Adding omitMandatoryCheck to create-mutation

ttheuerkauf:
See issue #203 for further details. Hoping to have implemented the flag as proposed.

### DIFF
--- a/src/GraphQL/Mutation/MutationType.php
+++ b/src/GraphQL/Mutation/MutationType.php
@@ -287,6 +287,7 @@ class MutationType extends ObjectType
                         'args' => [
                             'id' => ['type' => Type::nonNull(Type::int())],
                             'defaultLanguage' => ['type' => Type::string()],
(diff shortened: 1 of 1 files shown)


weisswurstkanone:
Mhm, I guess we need that for updates as well. What do you think ?

ttheuerkauf:
I was thinking about it and I am sure if it's needed. When creating objects, there is no chance to circumvent the problem because you cannot input data to two different languages. 

But when updating, the language can be set by defaultLanguage-Parameter. In this case it would be a semantical error in respect to the class definition. 

However, maybe it would be useful and should be added for consistency reasons?

weisswurstkanone:
Yes, definitely for completeness. Note that Pimcore respects the flags for updates so should do data-hub.as well

weisswurstkanone:
thanks a lot!

PR #211, opened 2020-04-27
add childrenSortBy and index to data-object and data-object-folder

dpfaffenbauer:


src/GraphQL/Resolver/DataObject.php
```diff
@@ -0,0 +1,76 @@
...
+ * @license    http://www.pimcore.org/license     GPLv3 and PEL
+ */
+
+namespace Pimcore\Bundle\DataHubBundle\GraphQL\Resolver;
+
+use GraphQL\Type\Definition\ResolveInfo;
+use Pimcore\Bundle\DataHubBundle\GraphQL\Service;
+use Pimcore\Bundle\DataHubBundle\GraphQL\Traits\ServiceTrait;
+
+class DataObject
```
weisswurstkanone, line 22: does it make sense to extend from Element resolver ?

src/GraphQL/Resolver/DataObject.php
```diff
@@ -0,0 +1,76 @@
...
+{
+    use ServiceTrait;
+
+    public function __construct(Service $graphQlService)
+    {
+        $this->setGraphQLService($graphQlService);
+    }
+
+    /**
+     * @param array            $value
```
weisswurstkanone, line 32: this is probably not an array (see how it is used in getById)

src/GraphQL/Resolver/DataObject.php
dpfaffenbauer: not sure honestly, if you want me to extend from it, I can change it

weisswurstkanone: I would vote for it, but this is probably a matter of taste

dpfaffenbauer: done

src/GraphQL/Resolver/DataObject.php
dpfaffenbauer: done

weisswurstkanone:
thanks a lot!

Issue #194, opened 2020-04-08
Error/Exception formatting


bramstroker:
When requesting an non existing object using datahub the following result is returned:

Query:
```
{
  getProduct(id: 15) {
    id
    sku
  }
}
```

Response
```
{
  "errors": [
    {
      "message": "Internal server error",
      "extensions": {
        "category": "internal"
      },
      "locations": [
        {
          "line": 2,
          "column": 3
        }
      ],
      "path": [
        "getProduct"
      ]
    }
  ],
  "data": {
    "getProduct": null
  }
}
```

Because of the generic message "Internal server error" our client cannot know what is wrong.

This behaviour is described in the Webonyx GraphQL documentation.
https://webonyx.github.io/graphql-php/error-handling/

The default behaviour is all exceptions not implementing `ClientAware` will be formatted as "internal server error".

I can register my own error formatter and display all full exception messages, but than I don't know if it is safe to display. So this is not really a solution.

In the Datahub bundle just a normal exception is thrown.
See: https://github.com/pimcore/data-hub/blob/master/src/GraphQL/Resolver/QueryType.php#L266
This exception is safe to display to the client, so the only right approach would be to throw a custom exception (ie `ClientSafeException`) implementing the `ClientAware` interface. This also applies to various other exceptions which are thrown in the datahub bundle.
Please let me know if you agree with this approach, than we can make a PR

weisswurstkanone:
Note that an exception can be thrown anywhere (there are other places in QueryType and trillion places outside of that class). We should at least implement this for other getters as well (put everything inside a try/catch block)
What would be the result in case ClientAware exception ?

Maybe have a look how this is handled in WebserviceController (could be an option to handle this at a central place).

            if (PIMCORE_DEBUG) {
                $debug = Debug::INCLUDE_DEBUG_MESSAGE | Debug::INCLUDE_TRACE | Debug::RETHROW_INTERNAL_EXCEPTIONS;
                $output = $result->toArray($debug);
            } else {
                $output = $result->toArray(false);
            }

As you see if pimcore is in debug mode then you will see a pretty error message.

            if (PIMCORE_DEBUG) {
                $debug = Debug::INCLUDE_DEBUG_MESSAGE | Debug::INCLUDE_TRACE | Debug::RETHROW_INTERNAL_EXCEPTIONS;
                $output = $result->toArray($debug);
            } else {
                $output = $result->toArray(false);
            }

bramstroker:
[...]

PR #212, opened 2020-04-28
Implement Client safe exceptions

bramstroker:
Here is a POC for the client safe exception as proposed in https://github.com/pimcore/data-hub/issues/194.
I only changed all exceptions which were thrown in `QueryType` for now. We can further refactor some other common cases which messages are perfectly safe and fine to show to the client.

Previously the error looked like this:

```
{
  "errors": [
    {
      "extensions": {
        "category": "internal"
      },
      "locations": [
        {
          "line": 2,
          "column": 3
        }
      ],
      "path": [
        "getProduct"
      ]
    }
  ],
  "data": {
    "getProduct": null
  }
}
```

In the new implementation the error message looks like this:

```
{
  "errors": [
    {
      "message": "object with ID 12 not found",
      "extensions": {
        "category": "pimcore.datahub"
      },
      "locations": [
        {
          "line": 2,
          "column": 3
        }
      ],
      "path": [
        "getProduct"
      ]
    }
  ],
  "data": {
    "getProduct": null
  }
}
```

When you have pimcore debug mode of the response has the following structure:

```
{
  "errors": [
    {
      "message": "object with ID 12 not found"
    }
  ]
}
```

Previously 

```
{
  "errors": [
    {
      "message": "Internal server error"
    }
  ]
}
```

Thoughts?

### DIFF
--- a/src/GraphQL/FieldHelper/DataObjectFieldHelper.php
+++ b/src/GraphQL/FieldHelper/DataObjectFieldHelper.php
@@ -19,6 +19,7 @@ use GraphQL\Language\AST\FieldNode;
 use GraphQL\Type\Definition\ResolveInfo;
 use GraphQL\Type\Definition\Type;
 use Pimcore\Bundle\DataHubBundle\GraphQL\DataObjectQueryFieldConfigGeneratorInterface;
+use Pimcore\Bundle\DataHubBundle\GraphQL\Exception\ClientSafeException;
 use Pimcore\File;
 use Pimcore\Logger;
 use Pimcore\Model\DataObject\ClassDefinition\Data;
@@ -206,7 +207,7 @@ class DataObjectFieldHelper extends AbstractFieldHelper
             case 'mutation':
                 return $this->getGraphQlService()->supportsDataObjectMutationDataType($typeName);
             default:
-                throw new \Exception("unknown operation type");
+                throw new ClientSafeException("unknown operation type " . $typeName);
         }
     }
 
--- a/src/GraphQL/AssetType/AssetType.php
+++ b/src/GraphQL/AssetType/AssetType.php
@@ -19,6 +19,7 @@ use GraphQL\Type\Definition\ObjectType;
 use GraphQL\Type\Definition\ResolveInfo;
 use GraphQL\Type\Definition\Type;
 use Pimcore\Bundle\DataHubBundle\GraphQL\ElementDescriptor;
+use Pimcore\Bundle\DataHubBundle\GraphQL\Exception\NotAllowedException;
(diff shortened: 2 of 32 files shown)


weisswurstkanone:
Nice! Can you add this to the other places as well? (maybe it is sufficient to cover everything where PimcoreDataHubBundle::getNotAllowedPolicy is involved).

In addition, can you check if the error messages can improved so that you get a clear picture what exactly is disallowed (could be a relation as well).

bramstroker:
I changed all occurrences of `throw` in the GraphQL namespace to throw a client safe exception.
Also introduced a new custom exception `NotAllowedException` to indicate the permission issues. This allows for more granular control (ie. catch only the not allowed cases).
The error messages seems fine to me.

What needs to be refactored is to eliminate all duplication of the isAllowed checks. All these lines are copy/pasted everywhere.
```
if (!WorkspaceHelper::isAllowed($element, $context['configuration'], 'read')) {
    if (PimcoreDataHubBundle::getNotAllowedPolicy() == PimcoreDataHubBundle::NOT_ALLOWED_POLICY_EXCEPTION) {
        throw new NotAllowedException('not allowed to view document');
    } else {
        return null;
    }
}
```

Better to introduce some method like `WorkspaceHelper::checkAllowed` which contains all this logic. Than only on one place the `NotAllowedException` is thrown and the exception message can be easily extended with more details without having to change several places.
Maybe you could create an issue for this in Github?

weisswurstkanone:
see #214

src/GraphQL/FieldHelper/DataObjectFieldHelper.php
```diff
@@ -206,7 +207,7 @@ public function supportsGraphQL(Data $fieldDefinition, string $operationType)
             case 'mutation':
                 return $this->getGraphQlService()->supportsDataObjectMutationDataType($typeName);
             default:
-                throw new \Exception("unknown operation type");
+                throw new ClientSafeException("unknown operation type " . $typeName);
```
weisswurstkanone, line 210: Not sure about this one (and probably a few others) as this indicates a coding error. What do you think?

bramstroker: Yes, you are right this exception should never happen in a normal case, only while coding new mutations/queries. However when it does happen it is no problem to show it to the client, so I don't see a good argument to change it back to a normal exception.

weisswurstkanone: Mhm, it will be still shown if pimcore is in debug mode

weisswurstkanone:
merci @bramstroker

PR #213, opened 2020-04-29
implement new event before loading listing data

juckerf:
part of the solution for #207

Issue #207, opened 2020-04-23
[Feature Request] Make it possible to load objects from specific hierarchy level


julmot:
When building a Product Listing Page we want to show items of the first hierachy level, which is the color variant in our case. Further nested levels like size etc. should not be displayed on the listing page, only on the detail page.

With our own PR #150 we've introduced the possibility to get children, so we could theoretically call something like a `getProductListing`, then directly go on the first variant hierarchy level and get all the information we need.

But this can not be a solution since it would not be possible to get a `totalCount` for this specific level nor would it work with paginations, where it should be possible to load only the first "X" elements and with an offset like specified in [the docs](https://github.com/pimcore/data-hub/blob/master/doc/graphl/querysamples/Sample_ManufacturerListing.md). Since it would not be relative to the specific variant level, but to the entire product listing.

The current solution would probably be to build a [custom query](https://github.com/pimcore/data-hub/blob/master/doc/graphl/AddCustomQuery.md), but then it'd be necessary to re-build default things like 

```
    defaultLanguage: $lang,
    sortOrder: "ASC",
    published: true,
    first: $limit,
    after: $offset
```

Or am I missing something?

Please take this as a Feature Request to load hierarchy specific objects.

juckerf:
probably it would help to have POST_BUILD event instead of just a PRE_BUILD (to extend behaviour of core types, resolvers etc.) or to have some way to inject some advanced custom filter logic in the "*Listing"-Endpoints (the current filter only allows to filter based on fields of the objects database table)

juckerf:
something like the legacy condition logic for webservices (https://pimcore.com/docs/6.x/Development_Documentation/Web_Services/Query_Filters.html#page_Legacy-Mode), just for data-hub, would allow one to add any custom filter logic to a query.
what do you think about it @weisswurstkanone ?

weisswurstkanone:
well, the code is already there and used for normal listings. see
https://github.com/pimcore/data-hub/blob/1a640d6574efc82708aeee69f0cedcd1fc353888/src/GraphQL/Helper.php#L19

juckerf:
If my understanding is correct this only allows very limited filter possibilities (based on simple equals, Like, gt, lt "where"-conditions on columns of the objects table).-- Sent from my Android phone with GMX Mail. Please excuse my brevity.On 28/04/2020, 09:25 Josef Aichhorn <notifications@github.com> wrote:

   
[...]

PR #216, opened 2020-04-30
add new post build events to mutation and query type config build

juckerf:
as discussed in #207 the new events in this PR provide the ability to alter the schema configuration before actually building the schema. E.g. for adding custom arguments to existing types (which can be particularly useful with the new feature of #213 )


weisswurstkanone:
merci @juckerf

PR #177, opened 2020-03-19
Add mediaqueries to AssetType

juckerf:
Adds a new field "srcset" to AssetType (solves issue #169)
Selection_019

PR #221, opened 2020-05-07
Add resolutions to AssetType

mugge6:
Adds new field "resolutions" to Assets and the "srcset" field.

Screenshot_1

- Types argument is optional (default: 2)
- Types argument can be a list [2, 4, 6]

### DIFF
--- a/src/GraphQL/AssetType/AssetType.php
+++ b/src/GraphQL/AssetType/AssetType.php
@@ -16,16 +16,10 @@
 namespace Pimcore\Bundle\DataHubBundle\GraphQL\AssetType;
 
 use GraphQL\Type\Definition\ObjectType;
-use GraphQL\Type\Definition\ResolveInfo;
 use GraphQL\Type\Definition\Type;
-use Pimcore\Bundle\DataHubBundle\GraphQL\ElementDescriptor;
-use Pimcore\Bundle\DataHubBundle\GraphQL\Exception\NotAllowedException;
 use Pimcore\Bundle\DataHubBundle\GraphQL\Service;
 use Pimcore\Bundle\DataHubBundle\GraphQL\Traits\ServiceTrait;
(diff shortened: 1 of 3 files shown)


julmot:
To everyone: We've introduced the general possibility to fetch media queries on asset objects, but the different resolutions were missing (#177). Now we're adding it with this PR for media queries and the asset object itself 🎉

weisswurstkanone:
thx @mugge6 👍

src/GraphQL/AssetType/AssetType.php
```diff
@@ -67,55 +61,61 @@ public function build(&$config)
         $propertyType = $this->getGraphQlService()->buildGeneralType('element_property');
         $elementResolver = new Resolver\Element('asset', $this->getGraphQlService());
 
+        $resolutionsType = Type::listOf(new ObjectType([
+            'name' => 'resolutions',
+            'fields' => [
+                'url' => Type::string(),
+                'resolution' => Type::int(),
```
NiklasBr, line 68: This can be an integer or a float: https://developer.mozilla.org/en-US/docs/Learn/HTML/Multimedia_and_embedding/Responsive_images#Resolution_switching_Same_size_different_resolutions

weisswurstkanone: changed it

src/GraphQL/AssetType/AssetType.php
```diff
@@ -67,55 +61,61 @@ public function build(&$config)
...
+            ],
+        ]));
+        $resolutionsArgumentsType = [
+            'type' => Type::listOf(Type::int()),
```
NiklasBr, line 72: I believe this must change as well @weisswurstkanone :)

weisswurstkanone: :-)

PR #239, opened 2020-06-15
Make workspace setting configurable

ckogler-elements:
Use-Case: Only show specific permissions

pimcore.plugin.pimcoreDataSyndicatorBundle.workspace.object = Class.create(pimcore.plugin.datahub.workspace.abstract, {

    type: "object",
    availableRights : ['read'],
    rightCheckboxHidden : true,
    rightCheckboxDisabled : true
});

### DIFF
--- a/src/Resources/public/js/workspace/abstract.js
+++ b/src/Resources/public/js/workspace/abstract.js
@@ -15,6 +15,8 @@
 pimcore.registerNS("pimcore.plugin.datahub.workspace.abstract");
 pimcore.plugin.datahub.workspace.abstract = Class.create({
 
+    availableRights : ["create", "read","update", "delete"],
+
     initialize: function (parent) {
         this.parent = parent;
         this.workspaces = this.parent.data.workspaces;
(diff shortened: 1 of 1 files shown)


src/Resources/public/js/workspace/abstract.js
```diff
@@ -34,12 +35,14 @@ pimcore.plugin.datahub.workspace.abstract = Class.create({
...
+        for (var i=0; i<this.availableRights.length; i++) {
 
             var checkConfig = {
-                text: t("plugin_pimcore_datahub_workspace_permission_" + availableRights[i]),
-                dataIndex: availableRights[i],
-                width: 70
+                text: t("plugin_pimcore_datahub_workspace_permission_" + this.availableRights[i]),
+                dataIndex: this.availableRights[i],
+                width: 70,
+                hidden: this.rightCheckboxHidden || false,
```
weisswurstkanone, line 44: ?

ckogler-elements: To hide the column. In my use case i have only the "read" permission which is allways checked and it should be hidden by default - and disabled...

weisswurstkanone:
please don't forget to sign the CLA

PR #244, opened 2020-07-08
respect language parameter for classification store queries

juckerf:
Overriding the language parameter with "default" seems like a typo to me(?)


weisswurstkanone:
thx!

Issue #243, opened 2020-07-06
Support sending apikey via headers in addition to sending it via query param


Iblis:
We would like to send the apikey via header, using either 'X-API-Key' or 'apikey' as the header key.
I can provide a PR for that if the feature is accepted.

weisswurstkanone:
I guess both ways should be supported.

Iblis:
Yes, I wrote this in a confusing way and will correct it: Not instead, but in addition to query param

fashxp:
but headers should be the preferred way (mentioned in docs etc)

weisswurstkanone:
@Iblis can you add a note please

weisswurstkanone:
@Iblis ping

Iblis:
Hi there! Sorry, got swamped the last days. I am still eager to work on this, but won't happen the next days. @weisswurstkanone should I add/prepare something here already?

weisswurstkanone:
Take your time! Thx :-)

weisswurstkanone:
implemented in #252

PR #252, opened 2020-07-27
Support sending apikey via headers in addition to sending it via query param

Iblis:
Implementation of #243

### DIFF
--- a/src/Configuration.php
+++ b/src/Configuration.php
@@ -175,7 +175,7 @@ class Configuration extends AbstractModel
         $this->configuration['general']['name'] = $this->name;
 
         $securityConfig = $this->getSecurityConfig();
-        if (($this->configuration['general']['active']  ?? false) && isset($securityConfig['method']) && $securityConfig['method'] === 'datahub_apikey') {
+        if (($this->configuration['general']['active']  ?? false) && isset($securityConfig['method']) && $securityConfig['method'] === self::CONFIG_NAME_APIKEY) {
             $apikey = $securityConfig['apikey'] ?? "";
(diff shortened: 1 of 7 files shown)


src/Service/CheckConsumerPermissionsService.php
```diff
@@ -0,0 +1,44 @@
...
+    /**
+     * @param Request $request
+     * @param Configuration $configuration
+     *
+     * @return void
+     */
+    public function performSecurityCheck(Request $request, Configuration $configuration): bool
+    {
+        $securityConfig = $configuration->getSecurityConfig();
+        if ($securityConfig['method'] === 'datahub_apikey') {
```
weisswurstkanone, line 32: I think we should store this constant in a central place. wdyt ?

Iblis: Good Idea. Do you have a class for Constants already? Or should we rather put it as a public const in Configuration.php, that is where it is used anyways, right?

weisswurstkanone:
No, there is no didicated class for this but I guess Configuration.php should also be fine.
Your choice :-)

src/Configuration.php
```diff
@@ -301,4 +301,6 @@ public function getSecurityConfig()
     {
         return $this->configuration['security'] ?? [];
     }
+
+    public const CONFIG_NAME_APIKEY = "datahub_apikey";
```
weisswurstkanone, line 305: please move this up and prefix it with the "topic". So something like SECURITYCONFIG_AUTH_APIKEY

weisswurstkanone:
thanks a lot!

PR #278, opened 2020-10-19
Provide response cache

nagarciah:
Draft for #273:
- config.yml file can be used to enable graphql output cache and set items lifetime. If not provided, defaults are: cache disabled, 30 seconds lifetime
- cache can be disabled for a single request (if debug mode is on) by setting query params to true (pimcore_nocache, pimcore_outputfilters_disabled)
- graphql output cache uses the default cache backend configured for pimcore (Doctrine, Redis,...)

### DIFF
--- a/src/Service/OutputCacheService.php
+++ b/src/Service/OutputCacheService.php
@@ -109,7 +109,7 @@ class OutputCacheService
         $input = json_decode($request->getContent(), true);
         $input = print_r($input, true);
         
-        return md5($clientname . $input);
+        return md5("output_" . $clientname . $input);
     }
     
     private function useCache(Request $request) : bool {
--- a/src/Service/OutputCacheService.php
+++ b/src/Service/OutputCacheService.php
@@ -15,10 +15,14 @@
 
 namespace Pimcore\Bundle\DataHubBundle\Service;
 
+use Pimcore\Bundle\DataHubBundle\Event\GraphQL\Model\OutputCachePreLoadEvent;
+use Pimcore\Bundle\DataHubBundle\Event\GraphQL\Model\OutputCachePreSaveEvent;
+use Pimcore\Bundle\DataHubBundle\Event\GraphQL\OutputCacheEvents;
 use Psr\Container\ContainerInterface;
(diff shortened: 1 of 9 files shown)


nagarciah:
@weisswurstkanone can you take a look at this proposal and let us know if you agree, please?

weisswurstkanone:
I believe we should also offer a "exclude pattern". What do you think ?

weisswurstkanone:
@nagarciah ping 🏓

nagarciah:
> 
> 
> @nagarciah ping 🏓

Hello! I like the idea of an "exclude pattern" and we have a story to work on it, just trying to find the time to do it, probably in the next weeks

weisswurstkanone:
Great! 👍

nagarciah:
Hello. I tried a filter with an array of regex to match against the target URL, however I think it is not to very useful since the target url of a request won't change unless the datahub-configuration name is changed, so probably it would be enough to filter based the full name of the configuration, without using regex. What do you think?

weisswurstkanone:
What about a server-side event listener. Then the listener can decide whether caching is allowed or not.

src/Service/OutputCacheService.php
```diff
@@ -0,0 +1,137 @@
...
+        \Pimcore\Cache::save($item, $key, $tags, $this->lifetime);
+    }
+    
+    private function computeKey(Request $request) : string {
+        $clientname = $request->get('clientname');
+        
+        $input = json_decode($request->getContent(), true);
+        $input = print_r($input, true);
+        
+        return md5($clientname . $input);
```
weisswurstkanone, line 112: IMHO, we should prefix this with something (just as we do it for object_, asset_, etc)

weisswurstkanone:
Could you also please mention this in the docs ?
https://github.com/pimcore/data-hub/blob/master/doc/10_GraphQL/README.md
and
https://github.com/pimcore/data-hub/blob/master/doc/10_GraphQL/10_Events.md

weisswurstkanone:
@nagarciah thanks a lot for your contribution!

PR #281, opened 2020-11-02
Use resolver for performance / fix missing quote

hlohaus:
Don't count, if i don't use it. :)


weisswurstkanone:
Could you please split this up into 2 PRs as the topics are not really related. Thx!

weisswurstkanone:
@hlohaus ping

hlohaus:
it wouldn't be mergeable?

PR #292, opened 2020-11-20
Editables

weisswurstkanone:


solverat:
@weisswurstkanone From the next version on, data-hub won't work with symfony 3.4 anymore. There is no pimcore 7 and the soonest stable version can be expected within 7.1. What I'm trying to say: consider release branches!

brusch:
Datahub is not yet finally released ... keep calm 😉

brusch:
... and it works with 6.8 resp. 6.9

PR #298, opened 2020-12-14
read query variables after preExecute event

juckerf:
this allows modification of the variable values (and therefore the whole query - if needed) in the preExecute event.
the change also introduces a backward compatible way to set the request property of the ExecutorEvent to an actual request object (like the event is created by it's constructor) and not only to a string


weisswurstkanone:
thx!

PR #303, opened 2020-12-28
Default language parameter in classification store queries

cesarab:
When the language is not indicated in a Classification Store query, the `defaultLanguage` (if present) should be taken into account before stating that the language should be set to `default`.

This modification will take into account the Locale of the GraphQL service. The Locale property contains the value of the `defaultLanguage` when this parameter is specified in the query.


weisswurstkanone:
thx!

PR #306, opened 2021-01-12
Update 10_Events.md

danmichaelo:
Changes:
- Recommending `autoconfigure` over manual tagging, since `autoconfigure` are used with new Symfony projects and seems to be best practice.
- Renaming the class to end with `Subscriber`, to make it more immediately clear what the class is.
- Adding link to Symfony to read more about the difference between Event Listeners and Subscribers, which can be useful for someone not too familiar with Symfony (like myself)


weisswurstkanone:
Thank you!

weisswurstkanone:
thx!

PR #314, opened 2021-01-26
Optimize config dialog

### LINKED ISSUE #250: [ConfigItem] Optimize for extending and adding additional adapters
extract items in `initialize` into seperate `getItems` method to allow customizing 
see https://github.com/pimcore/data-hub/blob/master/src/Resources/public/js/configuration/graphql/configItem.js#L33

extract save url to class variable to allow customizing
see https://github.com/pimcore/data-hub/blob/master/src/Resources/public/js/configuration/graphql/configItem.js#L539

### LINKED ISSUE #313: [Configuration] Dirty close doesn't close configuration

fashxp:
fixes #250 
fixes #313


weisswurstkanone:
Seems that the change detector is never started if there is only one configuration tab open. Maybe "onActivate" isn't called ?

fashxp:
> Seems that the change detector is never started if there is only one configuration tab open. Maybe "onActivate" isn't called ?

it is just called when switching between tabs. but changedetector setup at this place is wrong anyway, need to be in initialize (otherwise it would be setup with every tab switch). 

changed the content. ... is also linked to https://github.com/pimcore/pimcore/pull/8000

PR #317, opened 2021-02-01
Fixes pimcore/data-hub#316 localized blocks GraphQL

HageMaster3108:
Fixes pimcore/data-hub#316 by passing the requested language to the Block Getter in GraphQL Service.

Forgive me if this isn't state-of-the-art PHP, if not consider this a pointer into where the issue might originate. Been a while since I last did lots of PHP (back in the 5.x days!).


weisswurstkanone:
thx!

PR #325, opened 2021-03-02
[Workspaces] Use add instead of insert to add new store record

aarongerig:
This PR changes the UI behavior of adding new records to the workspace store. Instead of prepending new records this appends the newly created record. This makes it more intuitive for the user, which expects the record to be added at the bottom.


weisswurstkanone:
thx!

PR #341, opened 2021-04-08
Updated Webonyx to latest version and prepared file uploads via GraphQL

adriaans:
- Updated Webonyx to version ^14.6.x
- Made Pimcore data-hub compatible with Webonyx version ^14.6.x
- Added file upload service that transforms uploaed files in `Symfony\Component\HttpFoundation\File\UploadedFile`
- Added GraphQL type for uploads based on the implementation of https://github.com/Ecodev/graphql-upload

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -15,8 +15,8 @@
     ]
   },
   "require": {
-    "pimcore/pimcore": "^6.8 || ^10.0",
-    "webonyx/graphql-php": "^0.13"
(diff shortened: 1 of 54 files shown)


rubanooo:
We implemented this because uploading large files via base64 strings would cause the GraphQL lexer to be called for every character of the base64 string. For very small files this is not a very big problem, but when uploading files of multiple megabytes the request would take a really long time

weisswurstkanone:
Thanks a lot!

PR #350, opened 2021-05-05
Asset tags implementation v2

### LINKED ISSUE #345: implement asset tags
It would be great to have asset tags to be visible in response of GraphQL query. For example, in response to this query:

```graphql
{
  getAssetListing(ids:"4708") {
    totalCount
    edges {
      node {
        __typename
        ... on asset {
          filename
          type
          tags {
            id
            path
            name
          }
          metadata {
            name
          }
          properties {
            ... on property_text {
              name
            }
          }
        }
        ... on asset_folder {
          fullpath
          filename
        }
      }
    }
  }
}
```

to have something similar to this:
```json
{
  "data": {
    "getAssetListing": {
      "totalCount": 1,
      "edges": [
        {
          "node": {
            "__typename": "asset",
            "filename": "KA-2105_0000_unknown_1561.jpg",
            "type": "image",
            "tags": [
              {
                "id": "8",
                "path": "/Brand/KOVEA",
                "name": "KOVEA"
              }
            ],
            "metadata": null,
            "properties": []
          }
        }
      ]
    }
  }
}
```

ab-kily:
This pull request fixes #345 issue.


ab-kily:
@weisswurstkanone I've handled with CLA 😁

weisswurstkanone:
great, wdyt about adding this to the remaining 2 element types as well ? 🚀

ab-kily:
> great, wdyt about adding this to the remaining 2 element types as well ?

Well, I think it is not a problem, but I need to clarify one architectural thing, because I haven't got much experience with Pimcore  yet.

In my current implementation there is `AssetTag` class, so to implement it on *document* and *object* element types I need to create `DocumentTag` and `ObjectTag` classes. On the other hand, I'm not sure if there are significant differences in tag implementation for different element types, so the other way is to implement the only class `Tag` for all the element types. 

Do we need to have object-dependent tag classes or there should be the only tag type for all element types in GraphQL?

weisswurstkanone:
Not really necessary I would say from a schema perspective. Codewise I would even say let's put the implementation into a trait.

weisswurstkanone:
Second thought. Maybe change it to ElementTag. Implementation should be then the same for all types, right? wdyt

ab-kily:
Ok, thank you, I've got it. I'll try to make necessary fixes today.

ab-kily:
I've implemented tags functionality for all of 3 element types: *object*, *asset* and *document*. This is how json responses look like:
```json
{
  "data": {
    "getBrand": {
      "__typename": "object_Brand",
      "tags": [
        {
          "name": "BASK"
        }
      ]
    }
  }
}
```
```json
{
  "data": {
    "getDocument": {
      "__typename": "document_page",
      "tags": [
        {
          "name": "Пейзаж"
        }
      ]
    }
  }
}
```
```json
{
  "data": {
    "getAsset": {
      "__typename": "asset",
      "tags": [
        {
          "name": "KOVEA"
        }
      ]
    }
  }
}
```

weisswurstkanone:
Nice, could you please the conflicts 👍

ab-kily:
> Nice, could you please the conflicts

done )

weisswurstkanone:
Thanks a lot!

PR #351, opened 2021-05-07
Fix undefined notices

dvesh3:


weisswurstkanone:
please update the license info to PCL

Issue #285, opened 2020-11-06
How to disallow request on unpublished Objects


mugge6:
If you're requesting objects via GraphQL from a Frontend Application via the `get<ObjectName>(id: 4)` then the Object will be returned independent from the published status of the object. With a little programmer knowledge every one is able to request unpublished object data based on the request data visible in the browser console, with guessing the the object IDs.

For example: With this approach everyone can request new unpublished products from a product detail page as soon as he has at least one GraphQL request.

What is the best approach to avoid this sort of reverse engineering from a data dataset?

julmot:
One thing to note here: The `published` option that is available in listing queries is also defined client-side, which makes manipulable  to fetch data that is not intended to be public.

For the mentioned `get<ObjectName>(id: 4)` there's not even a way to specifiy a `published` filter client-side.

weisswurstkanone:
Would https://github.com/pimcore/data-hub/issues/226 be an option? The one view on this is about manipulating data and the other thing is fetch and render stuff. We have to combine both worlds :-) And probably in real life you need both.

weisswurstkanone:
@mugge6 ping

weisswurstkanone:
closed due to inactivity

Issue #214, opened 2020-04-30
Enforce workspace permission checks at a central place


weisswurstkanone:
What needs to be refactored is to eliminate all duplication of the isAllowed checks. All these lines are copy/pasted everywhere.
if (!WorkspaceHelper::isAllowed($element, $context['configuration'], 'read')) {
    if (PimcoreDataHubBundle::getNotAllowedPolicy() == PimcoreDataHubBundle::NOT_ALLOWED_POLICY_EXCEPTION) {
        throw new NotAllowedException('not allowed to view document');
    } else {
        return null;
    }
}

romainbessugesmeusy:
Hi, is it the reason we see a huge amount of queries like this:
`SELECT `read` FROM plugin_datahub_workspaces_object WHERE cid IN (1,1598,1215,1214,1213,24,20,11,10,1,1600) AND configuration = 'mg-app' AND `read`=1 ORDER BY LENGTH(cpath) DESC LIMIT 1` ?

PR #353, opened 2021-05-11
Add event to check custom permissions

zoidbergx:
From https://github.com/pimcore/data-hub/issues/285 we need to support some custom logic about permission in our case to check if an object is published or not.

After refactoring with https://github.com/pimcore/data-hub/issues/214 we could add an event to do our custom checks.

### DIFF
--- a/src/WorkspaceHelper.php
+++ b/src/WorkspaceHelper.php
@@ -241,7 +241,7 @@ class WorkspaceHelper
         /** @var EventDispatcher $eventDispatcher */
         $eventDispatcher = \Pimcore::getContainer()->get('event_dispatcher');
         $eventDispatcher->dispatch($event, PermissionEvents::PRE_CHECK);
-        if(!$event->isGranted()){
+        if (!$event->isGranted()) {
             throw new ClientSafeException('access for '.  $element->getFullPath() . ' denied');
         }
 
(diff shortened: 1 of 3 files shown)


src/WorkspaceHelper.php
```diff
@@ -234,6 +237,14 @@ public static function checkPermission($element, $type)
             return true;
         }
 
+        $event = new PermissionEvent($element, $type);
+        /** @var EventDispatcher $eventDispatcher */
+        $eventDispatcher = \Pimcore::getContainer()->get('event_dispatcher');
+        $eventDispatcher->dispatch($event, PermissionEvents::PRE_CHECK);
+        if (!$event->isGranted()) {
+            throw new ClientSafeException('access for '.  $element->getFullPath() . ' denied');
```
weisswurstkanone, line 245: IMHO, it only should throw an expection if the policy is NOT_ALLOWED_POLICY_EXCEPTION. see two lines below. wdyt?

weisswurstkanone:
@zoidbergx 🏓

weisswurstkanone:
@zoidbergx , last reminder . Hope we can get it in there :-)

src/WorkspaceHelper.php
zoidbergx: yes you're right

zoidbergx:
Sorry saw it late, your mention is added in the last commit

weisswurstkanone:
thanks a lot!!

PR #362, opened 2021-05-27
Element tags mutation

### LINKED ISSUE #361: implement GraphQL tags mutation
While there is an implementation to query tags for all objects types, it would be great to also have mutation implemented. Current schema doesn't support tags mutation:

image

I'm going to try to implement this for `createAsset` and `updateAsset` methods by providing a pull request.

ab-kily:
This pull request fixes #361 

Now it's possible to make the following mutation requests: 

```graphql
mutation test ($id:Int, $input: AssetInput) {
  
  updateAsset(id:$id, input: $input) {
    success
    message
    assetData {
      tags {
        id
        name
        path
      }
    }
  }
  
}
```

Variables:

```json
{
  "id": 7173,
  "input": {
    "tags": [
      {
        "id": "1"
      },
      {
        "path": "/Brand/BASK kids"
      }
    ]
  }
}
```

Response:

```json
{
  "data": {
    "updateAsset": {
      "success": true,
      "message": "asset updated: 7173",
      "assetData": {
        "tags": [
          {
            "id": "1",
            "name": "Brand",
            "path": "/Brand"
          },
          {
            "id": "2",
            "name": "BASK kids",
            "path": "/Brand/BASK kids"
          }
        ]
      }
    }
  }
}
```

And almost the same for `createAsset`.

### DIFF
--- a/src/GraphQL/Mutation/MutationType.php
+++ b/src/GraphQL/Mutation/MutationType.php
@@ -24,6 +24,7 @@ use Pimcore\Bundle\DataHubBundle\Configuration;
 use Pimcore\Bundle\DataHubBundle\Event\GraphQL\Model\MutationTypeEvent;
 use Pimcore\Bundle\DataHubBundle\Event\GraphQL\MutationEvents;
 use Pimcore\Bundle\DataHubBundle\GraphQL\FieldHelper\DataObjectFieldHelper;
+use Pimcore\Bundle\DataHubBundle\GraphQL\ElementTag;
 use Pimcore\Bundle\DataHubBundle\GraphQL\Service;
 use Pimcore\Bundle\DataHubBundle\GraphQL\Traits\ElementIdentificationTrait;
 use Pimcore\Bundle\DataHubBundle\GraphQL\Traits\ElementTagTrait;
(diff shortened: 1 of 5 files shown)


src/GraphQL/Mutation/MutationType.php
```diff
@@ -878,18 +881,34 @@ public function buildCreateAssetMutation(&$config, $context)
...
+                                        'success' => false,
+                                        'message' => 'no "id" nor "path" tag data defined for tag, or tag not found',
+                                    ];
+                                }
+                            } else {
+                                if ($key === 'data') {
+                                    $value = base64_decode($value);
+                                }
+                                $setter = 'set' . ucfirst($key);
+                                $newInstance->$setter($value);
```
ab-kily, line 959: BTW, to make the code more clean it would be great to have something like `Asset::setTags` implemented in pimcore/pimcore. If so, tags creation and update logic can be fully moved out of data-hub.

weisswurstkanone: yes, but please consider the minimum requirements:

  "require": {
    "pimcore/pimcore": "^6.9 || ^10.0",

ab-kily:
@weisswurstkanone I see my PR haven't passed php-cs-fixer test. But it seems there is something with the test. Should I react somehow?

weisswurstkanone:
Hi @ab-kily , wouldn't worry to much for now regarding cs fixer. But please could you have a look at the comment in the issue regarding documents and objects? Thanks a lot!

weisswurstkanone:
@ab-kily thanks a lot!
Just a thought. Maybe it would make sense to restrict access somehow. Wdyt ?

ab-kily:
> @ab-kily thanks a lot!
> Just a thought. Maybe it would make sense to restrict access somehow. Wdyt ?

Well, if you mean to restrict access to manipulation with tags I think it is a good idea. Maybe we need to add new items in Generic types block?

image

Something like "Asset tags", "Document tags" and so on...

weisswurstkanone:
Yes maybe, same for document "elements" for example. Let's give us some time to think about it. I just wanted to start a discussion :-) Maybe a configurator similar to the object schema definition would be an option. A flexible approach would be nice but it shouldn't make it overcomplicated to set it up.

PR #365, opened 2021-05-29
$value is not a instance of checkbox

archana-multi-concept:
I were having issue with BlockType.
when I request for query 
 ...on document_editableBlock{
          _editableName
          _editableType
          indices
        }
it return null value of editable name and type. but when removed condition of instance of checkbox. it started working. So I am proposing this change to plugin.
Because I don't want to change in my vendor folder. it will create problem while up-gradation. So If this change find correct then please merge.

### DIFF
--- a/src/GraphQL/DocumentElementType/BlockType.php
+++ b/src/GraphQL/DocumentElementType/BlockType.php
@@ -35,7 +35,7 @@ class BlockType extends ObjectType
                         '_editableName' => [
                             'type' => Type::string(),
                             'resolve' => static function ($value = null, $args = [], $context = [], ResolveInfo $resolveInfo = null) {
-                                if ($value instanceof Checkbox) {
+                                if ($value) {
                                     return $value->getName();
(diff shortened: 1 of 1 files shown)


src/GraphQL/DocumentElementType/BlockType.php
```diff
@@ -35,15 +35,15 @@ public static function getInstance()
                         '_editableName' => [
                             'type' => Type::string(),
                             'resolve' => static function ($value = null, $args = [], $context = [], ResolveInfo $resolveInfo = null) {
-                                if ($value instanceof Checkbox) {
```
weisswurstkanone, line 38: shouldn't it be 
$value instance of Pimcore\Model\Document\Editable\Block instead ?

Besides that, since this is a bug ... could you please rebase to 1.0 ?

weisswurstkanone:
Could you please clean up the imports and rebase on 1.0 branch please?

PR #366, opened 2021-05-31
Wrong $value instanceof checkbox

archana-multi-concept:
I have deleted old pull request mistakenly. So created new one. Let me know if any action required.


weisswurstkanone:
I guess

```
use Pimcore\Model\Document\Editable\Checkbox;
```

is not needed anymore in line 22

weisswurstkanone:
thx!

PR #374, opened 2021-06-14
fix rebuild definitions command

bramalho:
When running this command with a new version of PimcoreX (Symfony) we got the following error:
```
Return value of "Pimcore\Bundle\DataHubBundle\Command\GraphQL\RebuildDefinitionsCommand::execute()" must be of the type int, "null" returned.
```

This PR simply return `Symfony\Component\Console\Command\Command::SUCCESS` (`SUCCESS = 0`)


weisswurstkanone:
thx!

PR #377, opened 2021-06-17
Fix undefined array key "url" in AssetType resolver

mugge6:
Fix undefined array key "url" in AssetType resolver.


weisswurstkanone:
Hello @mugge6 , please check why the test fails. Thx!

PR #378, opened 2021-06-18
Fix DocumentTreeType 

dlhck:
duplicate type `_document_folder`

### DIFF
--- a/src/GraphQL/DocumentType/DocumentTreeType.php
+++ b/src/GraphQL/DocumentType/DocumentTreeType.php
@@ -63,8 +63,7 @@ class DocumentTreeType extends UnionType implements ContainerAwareInterface
             'document_hardlink',
             'document_link',
             'document_page',
-            'document_snippet',
-            '_document_folder'
+            'document_snippet'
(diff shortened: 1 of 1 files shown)


weisswurstkanone:
HI @davidhoeck , I see the issue :-) Does it cause a problem and how to reproduce. If so then this is rather something for the bugfix release. Thanks for your feedback!

dlhck:
@weisswurstkanone when using `*Listing` queries the error is happening, because the allowed types of a union type must be unique, which is not the case. so I think it is rather important to fix that

dlhck:
@weisswurstkanone any update on that?

weisswurstkanone:
Hi @davidhoeck , thanks for your feedback. Since this is a bug could you please rebase it to 1.0 ? Thanks a lot!

weisswurstkanone:
Hi @davidhoeck , fixed it directly via https://github.com/pimcore/data-hub/commit/f9c20decd07b69ebab42dd84232d44e36667ea51

Thanks for your input!

PR #379, opened 2021-06-21
Event dispatcher arguments have the wrong order

mugge6:
Dispatching method call from OutputCacheService has the wrong arguments order.


juckerf:
@weisswurstkanone the php-cs-fixer check fails during git fetch of `https://github.com/pimcore/data-hub`

dvesh3:
@mugge6 Since this is a fix, please rebase it on 10.0 branch. thanks!

PR #387, opened 2021-07-13
Allow thumbnail and asset dimensions to be queried

dpfaffenbauer:


dpfaffenbauer:
not sure why the tests fail...

src/GraphQL/Resolver/AssetType.php
```diff
@@ -250,6 +250,48 @@ public function resolveResolutions($value = null, $args = [], $context = [], Res
...
+     *
+     * @throws \Exception
+     */
+    public function resolveDimensions($value = null, $args = [], $context = [], ResolveInfo $resolveInfo = null)
+    {
+        if ($value instanceof ElementDescriptor) {
+            $thumbnailName = $args['thumbnail'] ?? null;
+
+            /**
+             * @var Asset\Image $asset
```
weisswurstkanone, line 269: Is it always of type image ?

dpfaffenbauer: it is for resolveResolution too.

weisswurstkanone:
What I wanted to ask ... for a text asset, what would be the result ? :-) Otherwise it looks good to me

dpfaffenbauer:
To be honest, I did it the same as `resolutions`, what happens there if it as text asset?

weisswurstkanone:
Hi @dpfaffenbauer , I would assume that it will crash :-) Can you give it a try please ?

dpfaffenbauer:
Yes, it does. Resolutions crashes as well as dimensions:

Bildschirmfoto 2021-07-14 um 13 12 09

weisswurstkanone:
See attached PR #388 which should fix this.

Works for 
{
  getAsset(id: 68) {
    filename
    resolutions(thumbnail: "exampleColor") {      
      url
    }
  }
}


but not for id 336 because 336 is not an image/video asset.

dpfaffenbauer:
changed it on my side

weisswurstkanone:
great, thanks!

PR #390, opened 2021-07-22
fix video data resolver

bramalho:
This PR fixes the issue to get the video data on `Video` types:

When run the query:
```
videoTest {
          type
          __typename
          data {
            __typename
            ... on asset {
              fullpath
            }
          }
        }
```

There was the error:
```
"message": "Warning: Undefined array key \"data\""
```

### DIFF
--- a/src/GraphQL/Resolver/Video.php
+++ b/src/GraphQL/Resolver/Video.php
@@ -129,7 +129,7 @@ class Video
                 $asset = $value->getData();
                 $fieldHelper = $this->getGraphQlService()->getAssetFieldHelper();
                 $fieldHelper->extractData($data, $asset, $args, $context, $resolveInfo);
-                $data['data'] = $data['data'] ? base64_encode($data['data']) : null;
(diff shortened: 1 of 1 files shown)


weisswurstkanone:
@bramalho I would classify this as a bug (should work with all error_reporting levels) .... could you please rebase this onto 1.0 instead of 1.x ? Thx!

src/GraphQL/Resolver/Video.php
```diff
@@ -129,7 +129,7 @@ public function resolveData($value = null, $args = [], $context = [], ResolveInf
                 $asset = $value->getData();
                 $fieldHelper = $this->getGraphQlService()->getAssetFieldHelper();
                 $fieldHelper->extractData($data, $asset, $args, $context, $resolveInfo);
-                $data['data'] = $data['data'] ? base64_encode($data['data']) : null;
+                $data['data'] = isset($data['data']) && !empty($data['data']) ? base64_encode($data['data']) : null;
```
ramundomario, line 132: `isset` is not necessary as `empty` already checks that the variable is set :wink:

ramundomario (changes requested):

PR #393, opened 2021-07-22
Add localizedfield support for all active languages

skoch98:
In API and GraphQL context, it would be nicer to directly be able to create or update multiple languages without complex and large mutations. 

I have to place one "Locale Switcher" per language and per field (depending on the applications this can be very much). Next, I need to enter a name and the language and finally move the field into it:
image
Query:
image

Using a small optimization, if you agree with it, is much easier:
image
Query:
image

New GraphQL-Definitions (generated for each Pimcore/GraphQL-Type that is referenced inside the operator):
image
image

Tested data-types was added to a whitelist: https://github.com/pimcore/data-hub/blob/41c279f32f007060c418450e211c4006e94e41fa/src/Resources/public/js/mutationoperator/LocaleCollector.js#L26


skoch98:
I would like to whitelist the supported fields in this PR using ExtJS.

skoch98:
> I would like to whitelist the supported fields in this PR using ExtJS.

Done since commit 41c279f32f007060c418450e211c4006e94e41fa

skoch-twocream:
> I would make this flexible in a way to allow to inject 3rd party data types
> > @skoch98 🏓 I believe it should be possible to add additional types here.

Whoops. I did not see your response! Sorry.

But additional types can be simple added by extending the javascript class?

Example:
```js
pimcore.registerNS("pimcore.plugin.datahub.mutationoperator.localecollector");

pimcore.plugin.datahub.mutationoperator.localecollector = Class.create(pimcore.plugin.datahub.mutationoperator.localecollector, {
    allowChild: function ($super, targetNode, dropNode) {   
        return $super(targetNode, dropNode) || in_array(dropNode.data.dataType, ["foo", "bar"]
    }
});
```

weisswurstkanone:
Not sure how this works if there are two extensions 🤔

skoch-twocream:
> Not sure how this works if there are two extensions 🤔

Where you see this?

Here? https://github.com/pimcore/data-hub/blob/1.x/src/DependencyInjection/Configuration.php#L35

weisswurstkanone:
thx!

PR #394, opened 2021-07-23
Add userId and omitVersionCreate to data-object mutation-type

skoch98:
If the GraphQL is used for automatic processes (e. g. in the ERP) it would be very helpful to identify in the version who the contact person is or which system updated the object.

Currently, no user is specified in the version:
image

To manage the user information in the data hub mask per each configuration does not seem to be flexible enough to me. So I implemented this via another field in the mutation:
image

Also in some cases there are several micro-processes running in an automatism for which reason it makes sense to suppress the creation of a version for specific use cases. Also (for example ERP processes) should not remove versions of real users through the version limit.
For this I added a field "omitVersionCreate". Or maybe you don't see it in this bundle? WDYT?

Both improvements are feedback I've received from customers at work.

### DIFF
--- a/src/GraphQL/Mutation/MutationType.php
+++ b/src/GraphQL/Mutation/MutationType.php
@@ -648,7 +648,7 @@ class MutationType extends ObjectType
 
                         $newInstance->save();
 
-                        if (isset($args['omitVersionCreate'])&& $args['omitVersionCreate'] && !$omitVersionCreateBefore) {
+                        if (isset($args['omitVersionCreate']) && $args['omitVersionCreate'] && !$omitVersionCreateBefore) {
                             Version::enable();
                         }
 
--- a/src/GraphQL/Mutation/MutationType.php
+++ b/src/GraphQL/Mutation/MutationType.php
@@ -40,6 +40,7 @@ use Pimcore\Model\DataObject\AbstractObject;
 use Pimcore\Model\DataObject\ClassDefinition;
 use Pimcore\Model\DataObject\Concrete;
 use Pimcore\Model\Document;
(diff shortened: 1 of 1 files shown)


weisswurstkanone:
Nice improvement!!! But would we need a similar stuff for  the "create" mutation as well? On top of it, let us do it for documents and assets as well.

skoch98:
> Nice improvement!!! But would we need a similar stuff for the "create" mutation as well? On top of it, let us do it for documents and assets as well.

Done ;-) Tested with following mutation types:
- createAsset
- createAssetFolder
- createDocumentEmail
- createDocumentLink
- createDocumentPage
- createObjectFolder
- createTable (Data-Object)
- updateAsset
- updateAssetFolder
- updateDocumentEmail
- updateDocumentLink
- updateDocumentPage
- updateObjectFolder
- updateTable (Data-Object)

To test my changes I had to fix several bundle bugs:
 - static method does not exist, see https://github.com/pimcore/data-hub/pull/394/commits/14a858ec737a20b844ed7e2d8f3d880448bc2fc9#diff-0f7b70d73908647c04c1eacf3d4951abd60b89f69fb0091af8c7816d01cc7bdaR273
 - `undefined method setHref`, see https://github.com/pimcore/data-hub/pull/394/commits/14a858ec737a20b844ed7e2d8f3d880448bc2fc9#diff-0f7b70d73908647c04c1eacf3d4951abd60b89f69fb0091af8c7816d01cc7bdaR387-R389

I also removed the `omitVersionCreate`-option on all create mutations.

weisswurstkanone:
@skoch98 Merci! 👍

PR #395, opened 2021-07-27
Fix warnings when adding operators to configuration

rubanooo:
When developing and adding operators to the config, the add join function fails because an operator does not have a attribute field defined. I do not know if this is the best solution to the problem, but it fixes the warnings and the query still runs just fine.


weisswurstkanone:
@rubanooo Thx!

PR #398, opened 2021-07-29
Add ability to list unpublished assets

longwave:
https://github.com/pimcore/data-hub/pull/12 allows unpublished objects to be listed, but this wasn't applied to assets.

Previously:
```
query { getAssetListing(published: false) { edges { node {  __typename } } } }
```
results in `Unknown argument "published" on field "getAssetListing" of type "Query".`

After this PR, this query works as expected.


weisswurstkanone:
https://github.com/pimcore/data-hub/blob/cd0a54724cfec20a901e9e584e15e030fe076a19/src/GraphQL/Resolver/QueryType.php#L418

weisswurstkanone:
Thanks @longwave 👍

PR #402, opened 2021-08-03
Update Helper.php

sissonen:
Columns may be operator type (in addition to attributes) such as thumbnail-operator. Added a check before going further since trying to access key 'attribute' results in an error here: $name = $attributes['attribute'];

### DIFF
--- a/src/GraphQL/Helper.php
+++ b/src/GraphQL/Helper.php
@@ -36,16 +36,19 @@ class Helper
         $parts = get_object_vars($filter);
         foreach ($parts as $key => $value) {
             foreach ($columns as $column) {
-                $attributes = $column['attributes'];
-                $name = $attributes['attribute'];
+                
+                if (!isset($column['isOperator']) or !$column['isOperator']) {
+                    $attributes = $column['attributes'];
+                    $name = $attributes['attribute'];
 
-                if (strpos($name, '~') !== false) {
-                    $nameParts = explode('~', $name);
-                    $brickName = $nameParts[0];
-                    $brickKey = $nameParts[1];
-                    if ($brickKey === $key) {
-                        $list->addObjectbrick($brickName);
-                        $mappingTable[$brickKey] = 1;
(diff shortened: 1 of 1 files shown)


weisswurstkanone:
@sissonen what be the steps to reproduce ? Besides that, could you please rebase the PR on 1.0? Thx!

sissonen:
I wasn't able to reproduce this on Pimcore demo, but on my local environment the problem arises thus:
- create a class (e.g. Product) with a gallery field
- create a graphQL end-point with data-hub with the Product in the Query schema
- in schema settings add Transformers -> Thumbnail to available fields and the gallery field under that + choose a thumbnail defition (e.g. grid)
- mark Asset -> Read as checked as well as required Assets and Objects folders in Security definition tab etc...

Without my update I get:
Screenshot_20210803_164933

With the update:
Screenshot_20210803_165050

The value of the $column variable just before the if-clause:
Screenshot_20210803_165552

 As opposed to "normal"/attribute column values:
Screenshot_20210803_165743

sissonen:
> @sissonen what be the steps to reproduce ? Besides that, could you please rebase the PR on 1.0? Thx!

This is my second ever PR... Any hints on how to "rebase the PR on 1.0"? :)

sissonen:
Ah, now that I pulled branch 1.0, it seems that this has been solved there with ([commit](https://github.com/pimcore/data-hub/commit/f662c1b136d86d41d87a83edff61aedc563fb5ec#diff-edef28c5331feef170a92ffb6e8e71e45a2f3a70898328d652e5130a9ca1a606)):
`if (isset($attributes['attribute'])) {`

weisswurstkanone:
Thanks for your feedback. Then this can be closed because it will be merged into 1.x on the next release day anyway - which will be tomorrow.

weisswurstkanone:
image
should do the trick.

Otherwise please checkout the 1.0 branch and submit a new PR in favor of this one.

PR #407, opened 2021-08-13
Use DataObject instead of AbstractObject, AbstractObject can't be instantiated in pimcore x anymore

Iblis:


weisswurstkanone:
merci!

PR #417, opened 2021-09-21
Implement save events

fashxp:
added `PRE_SAVE` and `POST_SAVE`


src/Event/ConfigurationEvents.php
```diff
@@ -28,4 +28,28 @@ final class ConfigurationEvents
...
+     *  - configuration | the original configuration instance
+     *
+     * @Event("Pimcore\Event\Model\GenericEvent")
+     *
+     * @var string
+     */
+    const CONFIGURATION_PRE_SAVE = 'pimcore.datahub.configuration.preSave';
+
+    /**
+     * Fired before a configuration was saved
```
weisswurstkanone:
```suggestion
     * Fired after a configuration was saved
```

Issue #421, opened 2021-10-11
[Improvement]: Use LocationAwareConfigRepository from pimcore/pimcore:^10.2


brusch:
### Improvement description

See pimcore/pimcore#9330

PR #424, opened 2021-10-22
Use LocationAwareConfigRepository from pimcore/pimcore:^10.2

mcop1:
Part of #421

### DIFF
--- a/doc/20_Deployment.md
+++ b/doc/20_Deployment.md
@@ -22,3 +22,56 @@ to do that for all definitions, or
 datahub:graphql:rebuild-definitions --definitions=newsapp,otherendpoint
 ```
 for specific definitions.
+
+Configuration Storage
+The configuration user interface utilizes the LocationAwareConfigRepository for storing the configuration. With the environment variable PIMCORE_WRITE_TARGET_DATA_HUB the storage location can be configured, possible values are
+
+symfony-config - write configs as Symfony Config as YAML files to /var/config/data-hub/<name>.yaml
+settings-store - write configs to the SettingsStore
+disabled - do not allow to edit/write configs at all
+Details also see Pimcore Docs.
+
+
+### Configuration Storage
+
+The configuration user interface utilizes the `LocationAwareConfigRepository` for storing the configuration. With
+the environment variable `PIMCORE_WRITE_TARGET_DATA_HUB` the storage location can be configured, possible values are
+- `symfony-config` - write configs as Symfony Config as YAML files to `/var/config/data-hub/<name>.yaml`
+- `settings-store` - write configs to the SettingsStore
+- `disabled` - do not allow to edit/write configs at all
+
+Details also see [Pimcore Docs](https://pimcore.com/docs/pimcore/10.2/Development_Documentation/Deployment/Configuration_Environments.html#page_Configuration-Storage-Locations-Fallbacks).
+
+#### Example
+```env 
+PIMCORE_WRITE_TARGET_DATA_HUB=symfony-config
+```
+
+Additionally, it is also possible to define the configuration directly in a symfony configuration file without using
+the configuration user interface. In this case, the configuration user interface is just read only.
+
+#### Sample Configuration File
+```yml 
+pimcore_data_hub:
+    configurations:
+        <name>:
+            general:
+                active: true
+                type: '<TYPE>'
+                name: '<NAME>'
+                description: '<DESCRIPTION>'
+                group: '<GROUP>
+                sqlObjectCondition: '<CONDITION>'
+                modificationDate: <DATE>
+                path: '<PATH>'
+                createDate: <DATE>'
+            schema:
+                queryEntities: {  }
+                mutationEntities: {  }
+                specialEntities:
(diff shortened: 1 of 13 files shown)


src/Configuration.php
```diff
@@ -95,6 +156,9 @@ public function setConfiguration($configuration): void
         if (is_array($configuration)) {
             $configuration = json_decode(json_encode($configuration), true);
         }
+        if (empty($this->getName()) == true) {
```
NiklasBr:
```suggestion
        if (empty($this->getName())) {
```
It's the same thing :)

src/Configuration/Dao.php
```diff
@@ -39,88 +40,146 @@ class Dao extends AbstractDao
...
     public static function getConfigModificationDate()
     {
-        $config = Config::locateConfigFile(self::CONFIG_FILE);
-
-        if (!file_exists($config)) {
-            return false;
-        }
-
-        return filemtime($config);
+        return strtotime('1987-10-09 13:58:00');
```
NiklasBr: This seems odd

src/Configuration.php
mcop1: Thanks for the suggestion and true- it is a bit of habit I developed over the years :)

fashxp:
please also add note to the docs ... similar to other bundles that are using the location aware config

src/Configuration/Dao.php
```diff
@@ -39,88 +40,146 @@ class Dao extends AbstractDao
...
     public static function getConfigModificationDate()
     {
-        $config = Config::locateConfigFile(self::CONFIG_FILE);
-
-        if (!file_exists($config)) {
-            return false;
-        }
-
-        return filemtime($config);
+        return strtotime('1970-01-01 00:00:00');
```
NiklasBr:
```suggestion
        return 0;
```

doc/20_Deployment.md
```diff
@@ -21,4 +21,48 @@ to do that for all definitions, or
...
+
+### Configuration Storage
+
+The configuration user interface utilizes the `LocationAwareConfigRepository` for storing the configuration. With
+the environment variable `PIMCORE_WRITE_TARGET_DATA_HUB` the storage location can be configured, possible values are
+- `symfony-config` - write configs as Symfony Config as YAML files to `/var/config/data-hub/<name>.yaml`
+- `settings-store` - write configs to the SettingsStore
+- `disabled` - do not allow to edit/write configs at all
+
+Details also see [Pimcore Docs](https://pimcore.com/docs/pimcore/10.2/Development_Documentation/Deployment/Configuration_Environments.html#page_Configuration-Storage-Locations-Fallbacks).
```
dvesh3:
```suggestion
Details also see [Pimcore Docs](https://pimcore.com/docs/pimcore/current/Development_Documentation/Deployment/Configuration_Environments.html#page_Configuration-Storage-Locations-Fallbacks).
```

src/Configuration/DatahubConfigLocator.php
```diff
@@ -0,0 +1,68 @@
...
+ *  @copyright  Copyright (c) Pimcore GmbH (http://www.pimcore.org)
+ *  @license    http://www.pimcore.org/license     GPLv3 and PCL
+ */
+
+namespace Pimcore\Bundle\DataHubBundle\Configuration;
+
+use Symfony\Component\Finder\Finder;
+
+/**
+ * Locates configs from path config/pimcore/headless-definitions.
```
dvesh3:
```suggestion
 * Locates data hub configs
```

src/Resources/public/js/configuration/graphql/configItem.js
```diff
@@ -15,11 +15,16 @@ pimcore.registerNS("pimcore.plugin.datahub.configuration.graphql.configItem");
...
     saveUrl: "/admin/pimcoredatahub/config/save",
+    disableForm: false,
 
     initialize: function (data, parent) {
         this.parent = parent;
         this.data = data.configuration;
         this.modificationDate = data.modificationDate;
+        if(data['configuration']['general']['writeable'] != null &&
+            data['configuration']['general']['writeable'] != undefined) {
+            this.disableForm = !data['configuration']['general']['writeable'];
```
dvesh3: writeable check should be negated here..right?

```suggestion
            this.disableForm = data['configuration']['general']['writeable'];
```

dvesh3: Nevermind...it is used in the disabled property so makes sense!

dvesh3: Ok...save button should consider this check at line 111

mcop1: When we disable the save button, the user would never get feedback on why he can´t save the configuration. Should we really do this? Wdyt?

dvesh3: We've done this for thumbnails configuration:
image

And it is better than making every field read only and then consistent with other features.

dvesh3 (approved):

PR #435, opened 2021-11-10
Fix Installer.php so that bundle properly reports as installed

NiklasBr:
`isInstalled()` would otherwise always return `false` since updating to 1.1.0

### DIFF
--- a/src/Installer.php
+++ b/src/Installer.php
@@ -81,4 +81,15 @@ class Installer extends AbstractInstaller
 
         return true;
     }
+
+    public function isInstalled()
+    {
+        try {
+            Db::get()->fetchOne('SELECT `cid` FROM `plugin_datahub_workspaces_document` LIMIT 1;');
+
+            return true;
+        } catch (\Exception $e) {
+            return false;
+        }
+    }
 }


dvesh3:
Installer switched to SettingsStoreAwareInstaller and this should set an entry for bundle is installed https://github.com/pimcore/data-hub/pull/432/files#diff-cabfa55630333e0f5e9327612d0d45d3dccec18765ae7d475a6d6d1c3afd3b36R45

NiklasBr:
This is what the logs say when updating from 1.0.9 and running the migrations:
```
15:33:36 [notice] Migrating up to Web2PrintToolsBundle\Migrations\PimcoreX\Version20210305134111
15:33:41 [notice] Migration Pimcore\Bundle\DataHubBundle\Migrations\PimcoreX\Version20210305134111 skipped during Pre-Checks. Reason: "Bundle not installed."
```

@dvesh3 looks like a catch-22 here, where the migration will not run unless isInstalled() returns true, but isInstalled() will not return true unless it is installed? How are we supposed to use DataHub here then?

NiklasBr:
Rolling back to 1.0.11 is another way to resolve it. If this is intended behaviour, how is DataHub supposed to be able to be installed in version 1.1.0? It is not even possible to manually install it in the admin UI in 1.1.0.

dvesh3:
@NiklasBr could you please check if PR #437 fixes the issue? thanks!

dvesh3:
closed in favor of #437

PR #441, opened 2021-11-23
fix: block create configuration if there is "-" in the name

RonaldSAMY:
if there is a '-' in the configuration name, it's crash while loading, so do the check in creation.


dvesh3:
@RonaldSAMY please sign CLA as this is your first contribution to this repo. thanks!

dvesh3:
@RonaldSAMY thanks for your PR! 

I believe restricting "-" for config keys is not an optimal fix. I have created a [PR](https://github.com/pimcore/data-hub/pull/444) to fix this issue by not normalizing config keys when using yaml storage. thanks!

PR #449, opened 2021-12-16
Setup PhpStan

blankse:

PR #450, opened 2021-12-17
Configuration: Fix wrong setName() in save() method

blankse:
Noticed in https://github.com/pimcore/data-hub/pull/449
Fix for 1.1 branch


weisswurstkanone:
thx!

PR #451, opened 2021-12-17
Fix usage of FieldcollectionDefinition and ObjectbrickDefinition

blankse:
Noticed in #449
Fix for 1.1 branch

- ObjectbrickDefinition extends FieldcollectionDefinition. So it must checked first.
- FieldcollectionDefinition has a getKey and not a getName method


weisswurstkanone:
thanks a lot!

PR #459, opened 2022-01-13
Querying on inputQuantityValue from Classificationstore returns null values (always)

### LINKED ISSUE #454: Querying on inputQuantityValue from Classificationstore returns null values (always)
### Expected behavior

After defining class attribute as a Classificationstore field (that can store i.e. "**select**", "**boolean**" and "**inputQuantiyValue**" types), a user should be able to obtain those fields using DataHub GraphQL API ( the _... on csFeature[type]_ methods). As Data Hub automatically generates GraphQL endpoints (and their documentation) to send queries against it is expected to fully consume them.

### Actual behavior

Unfortunately, whereas it is possible to execute queries against a  Classificationstore field of types "select" and "boolean", it is impossible to execute them against a  Classificationstore field of type "**inputQuantiyValue**" (other types haven't been thoroughly checked). The problem is that no matter their value (or unit) GraphQL query keeps returning **_null_** , see screen shot:
data_hub_issue_result
Object (product data sheet) and it's values:
data_hub_issue_definition

### Steps to reproduce

1. Create a Classificationstore than contains a group with relation to 
* boolean type
* select type
* **inputQuantityValue** type
data_hub_issue_calssstore

2. Make sure it's **inputQuantityValue** type
data_hub_issue_detail

3. Create a Class that contains an attribute with the Classificationstore you have created,
4. Create an instance of this Class (object, product data sheet) and fill it with some values,
5. Create GraphQL endpoint (Configuration) in Data Hub so your'e allowed to query against Classificationstore attribute,
6. Execute a query in iFrame (Data Hub) against object Classificationstore features, as in _Actual behavior_ - example:
[query....]   
features {
        ... on csFeatureBooleanSelect {
          type,
          name,
          description,
          checked
        }
        ... on csFeatureSelect {
          type,
          name,
          description,
          selection
        }
        ... on csFeatureInputQuantityValue {
        type,
        name,
        description,
        inputquantityvalue {
          value,
          unit {
            longname,
            abbreviation
          }
          toString
         }
        }
        ... on csFeatureInput {
          type,
          name,
          description,
          text
        }
      }

7.  You should see all values except those for **inpuQuantityValue** field

weisswurstkanone:
fixes  #454

### DIFF
--- a/src/GraphQL/Resolver/QuantityValue.php
+++ b/src/GraphQL/Resolver/QuantityValue.php
@@ -17,6 +17,7 @@ namespace Pimcore\Bundle\DataHubBundle\GraphQL\Resolver;
 
 use GraphQL\Type\Definition\ResolveInfo;
 use Pimcore\Bundle\DataHubBundle\GraphQL\Traits\ServiceTrait;
+use Pimcore\Model\DataObject\Data\InputQuantityValue;
 
 class QuantityValue
 {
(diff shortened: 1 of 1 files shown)


weisswurstkanone:
fixed via https://github.com/pimcore/data-hub/pull/459

PR #460, opened 2022-01-17
Fields not set when using create mutation

### LINKED ISSUE #452: Fields not set when using create mutation
### Expected behavior

When I create an object using a "create" mutation and provide parameters in the "input" section, I expect those parameters to be set when object is created.

### Actual behavior

The object is created with expected parent and key, but the values in the "input" section are ignored.

### Steps to reproduce

Create a simple object class ("testClass") with one text field, let's call it "testField". Then use the following GraphQL mutation:

`createTestClass (
    key: "testObject",
    parentId: 1,
    input: { testField: "testValue" }
) {
     message
}`

The object will be created, but with empty "testField".

weisswurstkanone:
fixes #452

### DIFF
--- a/src/WorkspaceHelper.php
+++ b/src/WorkspaceHelper.php
@@ -280,7 +280,7 @@ class WorkspaceHelper
                 $parent = $parent->getParent();
             }
         }
-        if (!is_null($element->getId())) {
+        if ($element->getId()) {
             $parentIds[] = $element->getId();
         }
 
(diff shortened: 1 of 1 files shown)


src/WorkspaceHelper.php
```diff
@@ -280,7 +280,9 @@ public static function isAllowed($element, Configuration $configuration, string
                 $parent = $parent->getParent();
             }
         }
-        $parentIds[] = $element->getId();
+        if (!is_null($element->getId())) {
```
blankse:
```suggestion
        if ($element->getId()) {
```
I think this should be enough, right? In older Pimcore Version it can be an 0 integer.

Issue #428, opened 2021-10-26
[Bug]: CORS error Pimcore 10 /w DataHub 1.0.9


snmabaur:
### Expected behavior

No CORS problem when fetching data from Pimcore 10 /DataHub 1.0.9

### Actual behavior

I have exactly the same setup with a Pimcore 6 and a Pimcore 10 and a Vue.js frontend... I fetch the data from localhost:3000 (frontend) to localhost:8080 with a JS fetch function.

On Pimcore 6 I get the data without problem on the endpoint //localhost:8080/pimcore-graphql-webservices/demo?apikey=XXXX but on Pimcore10 I get a CORS error

Access to fetch at 'http://localhost:8080/pimcore-graphql-webservices/hotspot?apikey=XXX' from origin 'http://localhost:3000' has been blocked by CORS policy: Response to preflight request doesn't pass access control check: No 'Access-Control-Allow-Origin' header is present on the requested resource. If an opaque response serves your needs, set the request's mode to 'no-cors' to fetch the resource with CORS disabled.

I tried to set the CORS headers in the nginx conf file but that also does not work. When I create a controller with a custom route then it works with the nginx headers but not for the datahub bundle

Could it be that the DataHub Bundle is not supporting Pimcore10?

Thanks

### Steps to reproduce

Install Pimcore 10 - Nuxt Frontend - Fetch data - CORS error

Install Pimcore 6 - Nuxt Frontend - Fetch data - no CORS problem

cmarcoin:
Could this be related to PHP 8.0 ?

In /vendor/pimcore/data-hub/src/Controller/WebserviceController.php line 159 we have : 

```php
        if (mb_stripos($contentType, 'multipart/form-data') !== false) {
            $input = $this->uploadService->parseUploadedFiles($request);
        } else {
            $input = json_decode($request->getContent(), true);
        }

        $query = $input['query'];
        $variableValues = isset($input['variables']) ? $input['variables'] : null;
```
But in the preflight request `$request->getContent() ` is empty so `$input` is null which leads to `$input['query']` throwing :

`request.CRITICAL: Uncaught PHP Exception ErrorException: "Warning: Trying to access array offset on value of type null"`

because of PHP 8.0

if you do something like 
```php
        $query = null;
        if (isset($input['query']))
            $query = $input['query'];
```
it works.

I have not found a workaround yet. If I do I will post it here.

FredoVelcro:
Hello,
[...]

PR #461, opened 2022-01-21
Fix: CORS error Pimcore 10 /w DataHub 1.0.9

blankse:
Fix https://github.com/pimcore/data-hub/issues/428


weisswurstkanone:
Thanks a lot!

PR #462, opened 2022-01-24
Escape configuration values properly

dvesh3:


weisswurstkanone:
let's discuss if there is a generic way. I assume that there other places in other bundles with grids like that

PR #463, opened 2022-01-24
PhpStan: Generate baseline on failure

blankse:


blankse:
@weisswurstkanone It works :)

weisswurstkanone:
Brilliant 🎉

PR #468, opened 2022-02-01
Fix for null blocks

dpfaffenbauer:


mcop1:
Thanks!

mcop1:
@dpfaffenbauer thanks, looks good, but since it is a bug, could you please change the target to 1.2?

dpfaffenbauer:
see https://github.com/pimcore/data-hub/pull/486

PR #472, opened 2022-02-11
Fix: InputQuantityValue Unit returns always empty array

### LINKED ISSUE #471: [Bug]: Bugfix FAILURE on Issue #454 - inputQuantityValue from Classificationstore returns null
### Expected behavior

From Issue [#454](https://github.com/pimcore/data-hub/issues/454):
After defining class attribute as a Classificationstore field (that can store i.e. "select", "boolean" and "inputQuantiyValue" types), a user should be able to obtain those fields using DataHub GraphQL API ( the ... on csFeature[type] methods). As Data Hub automatically generates GraphQL endpoints (and their documentation) to send queries against it is expected to fully consume them.

### Actual behavior

It seems that fix for the Issue [#454](https://github.com/pimcore/data-hub/issues/454) was only partly resolved as it returns now the _**value**_ of **_inputQuantityValue_** leaving the _**unit**_ object fileds null:

Moreover, taking a one-eye-look at the PR [#459](https://github.com/pimcore/data-hub/pull/459/files) files I cannot see any **_unit_** changes... 

### Steps to reproduce

The same as in the Issue [#454](https://github.com/pimcore/data-hub/issues/454):

Looking forward to heaving this feature ready as it super helpful!
B.R.
Andrzej

blankse:
Fix https://github.com/pimcore/data-hub/issues/471

### DIFF
--- a/src/GraphQL/Resolver/QuantityValue.php
+++ b/src/GraphQL/Resolver/QuantityValue.php
@@ -17,14 +17,14 @@ namespace Pimcore\Bundle\DataHubBundle\GraphQL\Resolver;
 
 use GraphQL\Type\Definition\ResolveInfo;
 use Pimcore\Bundle\DataHubBundle\GraphQL\Traits\ServiceTrait;
-use Pimcore\Model\DataObject\Data\InputQuantityValue;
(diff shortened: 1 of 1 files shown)


mcop1:
As always, thank you for your efforts

PR #474, opened 2022-02-14
[Bug] getCreateFolderResolver: changed abstract object to element service

### LINKED ISSUE #467: [Bug]: Internal Server Error on createObjectFolder with parent as path
### Expected behavior

When using createObjectFolder with path instead of parentId, the new object folder is created under the parent.

### Actual behavior

The server returns "Internal Server Error".

### Steps to reproduce

Query:

```
mutation {
  createObjectFolder(path: "/Example/Path", key: "NewFolderName") {
    success
    message
    id
  }
}
```

Returns:

```
{
  "errors": [
    {
      "extensions": {
        "category": "internal"
      },
      "locations": [
        {
          "line": 2,
          "column": 3
        }
      ],
      "path": [
        "createObjectFolder"
      ]
    }
  ],
  "data": {
    "createObjectFolder": null
  }
}
```

Datahub-Version: 1.2.0

After some investigations the reason seems to be the usage of `getByPath` on `AbstractObject` in file `src/GraphQL/Mutation/MutationType.php`:

grafik

In this function, a new instance is initalized with `new static()`, but this fails because `static` evaluates to `AbstractObject` and this is an abstract class:

grafik

Thank you very much :-)

kingjia90:
fixes #467

### DIFF
--- a/src/GraphQL/Mutation/MutationType.php
+++ b/src/GraphQL/Mutation/MutationType.php
@@ -1143,24 +1143,10 @@ class MutationType extends ObjectType
         return static function ($value, $args, $context, ResolveInfo $info) use ($elementType, $me) {
             $parent = null;
 
-            if ($elementType == 'asset') {
-                if (isset($args['parentId'])) {
-                    $parent = Asset::getById($args['parentId']);
-                } elseif (isset($args['path'])) {
-                    $parent = Asset::getByPath($args['path']);
-                }
-            } elseif ($elementType == 'document') {
-                if (isset($args['parentId'])) {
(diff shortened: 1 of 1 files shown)


blankse:
@kingjia90 The ElementService can also be used for documents and assets. So this lines can be dropped:
https://github.com/pimcore/data-hub/blob/44e042ef720e6a311dadd38d9508747e24d45619/src/GraphQL/Mutation/MutationType.php#L1146-L1158

kingjia90:
@blankse oh, you're right, thank you for the feedback!

PR #477, opened 2022-02-18
removes layout definition when saving the schema configuration

### LINKED ISSUE #464: [Bug] Store layout definition in schema configuration
### Feature description

why do we store layout definitions in the schema configuration of a datahub endpoint? 

image

kingjia90:
fixes #464

not sure if this is the best approach


fashxp:
@kingjia90 choose another way to remove that information ... please double check. thx.

PR #480, opened 2022-02-23
[Bug] Workspaces from non writeable configurations cannot be updated

### LINKED ISSUE #470: [Bug] Workspaces from non writeable configurations cannot be updated
### Improvement description

since configurations cannot be saved, also workspace settings are not updated in table.

mcop1:
resolves #470

### DIFF
--- a/src/Command/Configuration/RebuildWorkspacesCommand.php
+++ b/src/Command/Configuration/RebuildWorkspacesCommand.php
@@ -1,0 +1,93 @@
+<?php
+
+/**
+ * Pimcore
+ *
+ * This source file is available under two different licenses:
+ * - GNU General Public License version 3 (GPLv3)
+ * - Pimcore Commercial License (PCL)
+ * Full copyright and license information is available in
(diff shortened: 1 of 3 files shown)


src/Command/Configuration/RebuildWorkspacesCommand.php
```diff
@@ -0,0 +1,91 @@
...
+     *
+     * @throws \Exception
+     */
+    public function execute(InputInterface $input, OutputInterface $output)
+    {
+        $list = [];
+        $options = $input->getOption('configs');
+        if ($options) {
+            $option = $input->getOption('configs');
+            $configs = explode(',', $option);
```
kingjia90: everything seems working fine, but this $option and $options with same values, is there any reason for this?

probably it was meant to be like this:

```suggestion
        $options = $input->getOption('configs');
        if ($options) {
            $configs = explode(',', $options);
```

kingjia90 (changes requested):

PR #501, opened 2022-03-22
[Improvement]: disable introspection by default for production environment

### LINKED ISSUE #473: [Improvement]: disable introspection by default for production environment
### Improvement description

By default introspection for GraphQL is enabled, this applies to all environments. Because of the ease of setting up a GraphQL endpoint human errors can occur by accidentally adding sensitive data.

When this happens in production environments it is very easy to introspect the entire GraphQL endpoint and find out about this kind of data.

I think it would be a great improvement when introspection is disabled by default on production environments. Introspection could be enabled when you're logged into the Pimcore admin on a production environment and when you have the right permissions.

mcop1:
resolves #473


src/Migrations/Version20220322144548.php
```diff
@@ -0,0 +1,50 @@
...
+
+use Doctrine\DBAL\Schema\Schema;
+use Doctrine\Migrations\AbstractMigration;
+use Pimcore\Bundle\DataHubBundle\Configuration;
+
+final class Version20211108160248 extends AbstractMigration
+{
+    private function migrateConfiguration(bool $up)
+    {
+        $configs = Configuration::getList();
```
fashxp: this will deliver all configurations, not only graphql ones ... we need to filter for type graphql

fashxp (changes requested):

src/Migrations/Version20220322144548.php
```diff
@@ -0,0 +1,52 @@
...
+ *  @license    http://www.pimcore.org/license     GPLv3 and PCL
+ */
+
+namespace Pimcore\Bundle\DataHubBundle\Migrations;
+
+use Doctrine\DBAL\Schema\Schema;
+use Doctrine\Migrations\AbstractMigration;
+use Pimcore\Bundle\DataHubBundle\Configuration;
+
+final class Version20211108160248 extends AbstractMigration
```
fashxp: - migration actually needs to be located in  Pimcore\Bundle\DataHubBundle\Migrations\PimcoreX namespace
- filename and classname need to match
- should extend from `BundleAwareMigration`  ... to make sure it is not executed when bundle is not installed

fashxp (changes requested):

fashxp (changes requested):
migration cannot update the configs, if they are yaml files located in custom locations like bundles etc. 

I guess we need to get rid of the migration and introduce a BC layer ... if config value not set, set it to true.

src/Migrations/Version20220322144548.php
mcop1: > * migration actually needs to be located in  Pimcore\Bundle\DataHubBundle\Migrations\PimcoreX namespace
> * filename and classname need to match
> * should extend from `BundleAwareMigration`  ... to make sure it is not executed when bundle is not installed

As discussed, I changed the checkbox from "Enable Introspection" to "Disable Introspection". Default: Introspection is enabled.

mcop1:
> migration cannot update the configs, if they are yaml files located in custom locations like bundles etc.
> 
> I guess we need to get rid of the migration and introduce a BC layer ... if config value not set, set it to true.

As discussed, I changed the checkbox from "Enable Introspection" to "Disable Introspection". Default: Introspection is enabled. No more migration needed.

fashxp (approved):

PR #507, opened 2022-04-04
[Feature]: GraphQL Mutation for Advanced Many-to-Many Object relation

### LINKED ISSUE #491: [Feature]: GraphQL Mutation for Advanced Many-to-Many Object relation
### Feature description

Thers actual no way to save an Advanced Many-to-Many Object relation with the graphql api.

mcop1:
Resolves #491


fashxp (changes requested):
this actually implements many-to-many relation, but not many-to-many object relation
and for that, we cannot set any metadata, this we should add there too.

mcop1:
> this actually implements many-to-many relation, but not many-to-many object relation and for that, we cannot set any metadata, this we should add there too.

I added many-to-many-object relations too. Including the functionality to add values to custom columns (metadata).

fashxp (approved):

PR #520, opened 2022-04-21
Bugfix/519 fix block undefined array key

### LINKED ISSUE #519: [Bug]: Undefined array index not checked in Block Fields
### Expected behavior

In Debug mode, retrieving a block field value for a block field which never had a value, I'd expect it to return null (or an empty string).

### Actual behavior

Instead, it gives me `Warning: undefined array key "fieldname"`.

### Steps to reproduce

I have a query
```{
  getOrthomolCorporateProductDosageFormListing(defaultLanguage: "de_DE") {
   edges {
     node {
       id
       name(language: "de_DE")
       packsizes {
         size
         PZN
         EAN
         oNumber
         snapnextId
         b2cAppPoints
         packshot { # this is an asset field
           fullpath
           srcset(thumbnail: "PictureText") {
             url
             descriptor
           }
         }
       }
     }
   }
  }
}
```

In this model `OrthomolCorporateProductDosageForm` has a block `packsizes`. If I have one of these fields unset (never had a value) the serialized data in the db for that block is missing the array key (e.g. `packshot`).

If now I run pimcore in debug mode, datahub tries to access the array key nevertheless in [BlockEntryType.php Line 136](https://github.com/pimcore/data-hub/blob/v1.0.11/src/GraphQL/DataObjectType/BlockEntryType.php#L136)

So what's missing there is
```
if (!array_key_exists($resolveInfo->fieldName, $value)) {
    return null;
}
```

I'd have sent a PR right a way but currently github says "no more forks can be created" for this repo?

HageMaster3108:
Fixes #519

### DIFF
--- a/src/GraphQL/DataObjectType/BlockEntryType.php
+++ b/src/GraphQL/DataObjectType/BlockEntryType.php
@@ -129,6 +129,10 @@ class BlockEntryType extends ObjectType implements ContainerAwareInterface
             if (!is_array($value)) {
                 return null;
             }
+            
+            if (!array_key_exists($resolveInfo->fieldName, $value)) {
+               return null;
+            }
 
             $value = $value[$resolveInfo->fieldName];
 
--- a/src/GraphQL/DataObjectType/BlockEntryType.php
+++ b/src/GraphQL/DataObjectType/BlockEntryType.php
@@ -129,9 +129,9 @@ class BlockEntryType extends ObjectType implements ContainerAwareInterface
             if (!is_array($value)) {
                 return null;
             }
-            
+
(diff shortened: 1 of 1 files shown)


mcop1:
@HageMaster3108 Sorry, we had some troubles with the branches, could you please rebase the pr to the 1.3 branch? Thank you for your efforts.

HageMaster3108:
Sorry for the mess in my rebase operation, somehow had a commit in which didn't belong here and struggled to get rid of it ;-) Typically/almost never in a situation where I need to rebase :-) Hope it's ok now.

mcop1:
No problem, looks good now, thank you :)

PR #526, opened 2022-05-02
Removed PHPStan warning, added automated tests

mcop1:
- substituted query() with executyQuery() in Installer.php to get rid of the phpstan error
- added automated test via docker 
- added cache_router: false, because we don´t need it for testing

### DIFF
--- a/phpstan.neon
+++ b/phpstan.neon
@@ -1,5 +1,6 @@
 parameters:
     level: 4
+    reportUnmatchedIgnoredErrors: false
     paths:
         - src
     bootstrapFiles:
--- a/phpstan.neon
+++ b/phpstan.neon
@@ -1,7 +1,5 @@
 parameters:
     level: 4
-    ignoreErrors:
-        - '~^Method Doctrine\\DBAL(\\.*)?Connection::query\(\) invoked with \d+ parameters?, 0 required\.\z~'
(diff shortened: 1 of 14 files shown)


phpstan.neon
```diff
@@ -1,5 +1,6 @@
 parameters:
     level: 4
+    reportUnmatchedIgnoredErrors: false
```
blankse: You can ignore it like in the pimcore repository:
```suggestion
    ignoreErrors:
        - '~^Method Doctrine\\DBAL(\\.*)?Connection::query\(\) invoked with \d+ parameters?, 0 required\.\z~'
```

blankse: Why does this not reported in the lowest dependency... Can you increase the phpstan version in the composer? So both tests are with the newest/equal phpstan version?

blankse: Maybe we can use executeQuery() instead of query(). query() is deprecated in dbal:
https://github.com/doctrine/dbal/blob/dc9b3c3c8592c935a6e590441f9abc0f9eba335b/lib/Doctrine/DBAL/Connection.php#L1431

mcop1: I substituted query() with executyQuery() and removed the error warning pattern, thanks for the hint.

PR #554, opened 2022-06-29
Fix for phpstan error messages

mcop1:
Fix for error messages in lowest requirement configuration. There were some changes to the return/parameter types with 10.4.4.


mcop1:
Cherry picked into 1.4, therefore added to milestone 1.4.2

PR #573, opened 2022-07-22
[Feature]: Implement "getTranslation" and filter for languages

### LINKED ISSUE #557: [Feature]: Implement "getTranslation" and filter for languages
### Feature description

As soon as https://github.com/pimcore/pimcore/pull/12539 was merged, we can use it to implement a filter for one or more languages. 

- [x] Use new filter option in GraphQL fo getTranslation and getTranslationListing
- [x] Implement "getTranslation" method with existing filter options

robertSt7:
resolve #557


composer.json
```diff
@@ -12,7 +12,7 @@
   },
   "require": {
     "php": "^8.0",
-    "pimcore/pimcore": "^10.2",
+    "pimcore/pimcore": "^10.5",
```
mcop1, line 15: Maybe we should check if the getLanguages method is available or not and exclude/include it from/to the schema?

mcop1: Just saw the other pr, nevermind, since we have to bump it to 10.5 anyways (because of the plugins)

PR #574, opened 2022-07-26
FQN for Type

yariksheptykin:
The example mixes FQN and shortcuts for `Type`. I suggest we use FQNs everywhere.

### DIFF
--- a/doc/10_GraphQL/04_Query/20_Add_Custom_Query.md
+++ b/doc/10_GraphQL/04_Query/20_Add_Custom_Query.md
@@ -31,7 +31,7 @@ if you need more information on Pimcore's event mechanism.
 
         $operation = [
             'type' => $outputType,
-            'args' => ['itemId' => ['type' => Type::nonNull(Type::int())]],
+            'args' => ['itemId' => ['type' => \GraphQL\Type\Definition\Type::nonNull(Type::int())]],
             'resolve' => function ($source, $args, $context, \GraphQL\Type\Definition\ResolveInfo $info) {
                 // resolve the item using the input parameters. Result will be passed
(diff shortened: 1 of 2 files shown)


mcop1:
@yariksheptykin Thanks for the pr, could you please add it to these 2 files too?

https://github.com/pimcore/data-hub/blob/1.x/doc/10_GraphQL/07_Mutation/27_Add_Custom_Mutations.md
https://github.com/pimcore/data-hub/blob/1.x/doc/10_GraphQL/10_Events.md

yariksheptykin:
> @yariksheptykin Thanks for the pr, could you please add it to these 2 files too?
> 
> https://github.com/pimcore/data-hub/blob/1.x/doc/10_GraphQL/07_Mutation/27_Add_Custom_Mutations.md https://github.com/pimcore/data-hub/blob/1.x/doc/10_GraphQL/10_Events.md

Good idea, @mcop1 . I've updated the mutation example.

The events-example is o.k. because we import `Type` with `use` there. So I someone simply copy-pastes this example it is going to work. This was not the case with the other two places.

mcop1:
> > @yariksheptykin Thanks for the pr, could you please add it to these 2 files too?
> > https://github.com/pimcore/data-hub/blob/1.x/doc/10_GraphQL/07_Mutation/27_Add_Custom_Mutations.md https://github.com/pimcore/data-hub/blob/1.x/doc/10_GraphQL/10_Events.md
> 
> Good idea, @mcop1 . I've updated the mutation example.
> 
> The events-example is o.k. because we import `Type` with `use` there. So I someone simply copy-pastes this example it is going to work. This was not the case with the other two places.

Great, you are right, thank you!

PR #575, opened 2022-07-27
[Bug]: Changing Data-Hub display of fields to Names instead of Titles

### LINKED ISSUE #567: [Bug]: Changing Data-Hub display of fields to Names instead of Titles
### Improvement description

Apparently, there has been a change of being able to add single fields of Object Bricks instead of the whole Brick. Which is like a nice feature. 

However, I have been struggling a rather specific problem. 

1.) If you do not have a title of a field, which I do mainly for aesthetic reasons, so that two fields are right next to each other. The name does not properly displayed in the Schema after saving, but, instead, is shown as "Root". The good news is, that the functionality of fetching for it in GraphQL is not negatively impacted by it and still possible. 

root_datahub

image

image

image

The aesthetic reason why I went for no title, can be seen here (in order to display :

image

Therefore, I am not sure whether it would be better to switch to display simply the field names in order to avoid this bug, especially since one has to fetch them by the name and not the title anyways. There might be another workaround, with which I am not familiar with.

robertSt7:
resolve #567

### DIFF
--- a/src/Resources/public/js/fieldConfigDialog.js
+++ b/src/Resources/public/js/fieldConfigDialog.js
@@ -242,7 +242,7 @@ pimcore.plugin.datahub.fieldConfigDialog = Class.create({
                     child = child[0];
                 } else {
                     var attributes = nodeConf.attributes;
-                    var text = ts(attributes.label);
+                    let text = ts(attributes.label) ? ts(attributes.label) : `(${attributes.attribute})`;
 
                     if (attributes.dataType !== "system" && this.showFieldname && attributes.key) {
                         text = text + " (" + attributes.key.replace("~", ".") + ")";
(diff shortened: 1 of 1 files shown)


src/Resources/public/js/fieldConfigDialog.js
```diff
@@ -242,7 +242,7 @@ pimcore.plugin.datahub.fieldConfigDialog = Class.create({
                     child = child[0];
                 } else {
                     var attributes = nodeConf.attributes;
-                    var text = ts(attributes.label);
+                    let text = ts(attributes.label) ? ts(attributes.label) : `(${attributes.attribute})`;
```
blankse: ts() is deprecated
```suggestion
                    let text = attributes.label ? t(attributes.label) : `(${attributes.attribute})`;
```

blankse: See https://github.com/pimcore/pimcore/pull/12771

PR #578, opened 2022-07-28
[Bug]: Editable Many-To-One Relation returns 'null' for attributes of related object always

### LINKED ISSUE #555: [Bug]: Editable Many-To-One Relation returns 'null' for attributes of related object always
### Expected behavior

When querying for a related object via editable of type **Many-To-One Relation** (singular) I expect populated fields/values of this object to be returned in the response. 

### Actual behavior

Now it seems like attributes/fields of the related object (here: car) are **returned as 'null' always**, when using type **Many-To-One Relation** …

image

Same query using type Many-to-Many Relation instead returns expected value of attribute like shown below …

image

## Reproduce 'null' value
1. In case of Headless Documents add new Editable of type **Many-To-One Relation** for objects of type car to the template.
2. Ensure to provide Car Entity for Query Schema with at least attribute **name** to be populated.
3. Set/Configure **Datahub Endpoint** to allow access to Documents and Objects.
4. Query for Document where you have set relation to specific car and try to fetch field **name** (see section _Actual behavior_ above).   

image

## To fetch value of field 'name'

1. Change same editable to type **Many-To-Many Relation**.
2. Relate corresponding Car object once more.
3. Query for Document where you have set relation to specific car and try to fetch field **name**. Keep in mind to query for "document_editableRelations" and iterate on "relations" (plural) now (see section _Actual behavior_ above).

image

mcop1:
Resolves #555

### DIFF
--- a/src/GraphQL/DocumentElementType/RelationType.php
+++ b/src/GraphQL/DocumentElementType/RelationType.php
@@ -84,12 +84,12 @@ class RelationType extends ObjectType
                         ],
                         'relation' => [
                             'type' => $anyTargetType,
-                            'resolve' => static function ($value = null, $args = [], $context = [], ResolveInfo $resolveInfo = null) {
+                            'resolve' => static function ($value = null, $args = [], $context = [], ResolveInfo $resolveInfo = null) use ($graphQlService) {
                                 if ($value instanceof Relation) {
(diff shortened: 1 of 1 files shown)


neusta-diaspero:
Thank you for fixing that.

PR #579, opened 2022-07-28
[Improvement]: Replace plugins with event listener

### LINKED ISSUE #566: [Improvement]: Replace plugins with event listener
### Improvement description

https://github.com/pimcore/pimcore/pull/12633/files#diff-736172470d63f25a8d80e1457a5f8ef0ce9af5d798239430886fe1a410842aa3

Replace pimcore.plugin.broker.fireEvent with dispatchEvent and make events compatible for plugins like: 
https://github.com/pimcore/perspective-editor/pull/81

Change dependency from pimcore/pimcore to version 10.5

robertSt7:
resolve #566

### DIFF
--- a/composer.json
+++ b/composer.json
@@ -12,7 +12,7 @@
   },
   "require": {
     "php": "^8.0",
-    "pimcore/pimcore": "^10.2",
+    "pimcore/pimcore": "^10.5",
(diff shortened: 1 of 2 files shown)


blankse:
@robertSt7 You need to bump Pimcore Version to 10.5, right?

PR #581, opened 2022-08-04
Replace deprecated ts() function with t()

blankse:
ts() is deprecated since Pimcore 6.5.2 https://github.com/pimcore/pimcore/pull/5840
Throws console errors since Pimcore 10.5.0 https://github.com/pimcore/pimcore/pull/12771
Will be removed in Pimcore 11.0.0 https://github.com/pimcore/pimcore/pull/12826


mcop1:
Thanks!

PR #582, opened 2022-08-04
Fix: PhpStan-baseline error

robertSt7:


blankse:
@robertSt7 ~~Seems you need bump the phpstan version to "^1.8.2". So that the lowest test also use the new phpstan version.~~

Edit: The lowest Pimcore version has a wrong return type for getPosterAsset. I think we could fix it with an inline phpdoc
```
/** Image|null $relation */
$relation = $value->getPosterAsset();
```

src/GraphQL/DocumentElementType/VideoType.php
```diff
@@ -85,6 +85,7 @@ public static function getInstance(Service $graphQlService, AssetType $assetType
                             'type' => $assetType,
                             'resolve' => static function ($value = null, $args = [], $context = [], ResolveInfo $resolveInfo = null) use ($graphQlService) {
                                 if ($value instanceof Video) {
+                                    /** Image|null $relation */
```
blankse:
```suggestion
                                    /** @var Image|null $relation */
```
Exist the use statement of the Image class?

PR #587, opened 2022-08-12
Fix: Error when resolutions are used with asset folders

blankse:
We have a relation which can have folders or images. If I want the resolutions from the images I get an error:
image


mcop1:
Thank you 👍

PR #592, opened 2022-08-19
Fix: FolderTypes have wrong modificationDate field

blankse:
Before:
image
After:
image

modificationDateDate is always null


mcop1:
Thanks!

PR #593, opened 2022-08-19
Fix: Missing creationDate/modificationDate on Pimcore Object Type

blankse:
All element types have creationDate/modificationDate fields. It's only missing on the concrete objects.


mcop1:
And another one: Thank you!

PR #608, opened 2022-09-19
[DOCS] - Update "Customizing the endpoint" page

SamyMP:
The docs to customize an endpoint in the data-hub was not up to date. This should be the new config to achieve this


mcop1:
Thanks!

PR #633, opened 2022-11-17
[Bug] Fix Undefined array key Warning when a mutation has no column config yet

ultramegatom:
Fix Warning: Warning: Undefined array key "columns"
Fix Warning: Undefined array key "columnConfig"


mcop1:
Thank you!

PR #653, opened 2022-12-14
Fix: php stan

robertSt7:


blankse:
@robertSt7 I think we can simply remove the if in line 306, because one of the variables $isIdSet and $isFullpathSet must be true. See line 275.

So $conditionParts can't be empty.

PR #663, opened 2023-01-13
[Task]: Remove `pimcore:system:requirements:check`

kingjia90:
Once https://github.com/pimcore/pimcore/issues/12519 is done in 11.x, this command will be not available out of the box, therefore opening this PR as reminder.
If it is still necessary, please change the PR to add the bundle to require-dev.


fashxp:
resolved in https://github.com/pimcore/data-hub/pull/661 ... not really necessary.

PR #668, opened 2023-01-19
Attempt to Query by Tag Names Via DataHub

### LINKED ISSUE #611: [Feature]: Being able to filter by tags
### Feature description

Right now data-hub allows filtering on queries using Pimcore's webservice filter logic.
I was not able to find a way to use this to filter for tags.
Looking at https://pimcore.com/docs/pimcore/current/Development_Documentation/Best_Practice/Using_Tags_for_Filtering.html it seems to be straightforward to implement explicit tag-filters for queries.
Not sure though how to allow configuration for options like "considerChildTags" true/false

mazorg:
resolves #611

### DIFF
--- a/src/GraphQL/Resolver/QueryType.php
+++ b/src/GraphQL/Resolver/QueryType.php
@@ -9,8 +9,8 @@
  * Full copyright and license information is available in
  * LICENSE.md which is distributed with this source code.
  *
- *  @copyright  Copyright (c) Pimcore GmbH (http://www.pimcore.org)
- *  @license    http://www.pimcore.org/license     GPLv3 and PCL
+ * @copyright  Copyright (c) Pimcore GmbH (http://www.pimcore.org)
+ * @license    http://www.pimcore.org/license     GPLv3 and PCL
  */
 
 namespace Pimcore\Bundle\DataHubBundle\GraphQL\Resolver;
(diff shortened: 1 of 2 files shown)


src/GraphQL/Resolver/QueryType.php
```diff
@@ -297,6 +298,17 @@ public function resolveObjectGetter($value = null, $args = [], $context = [], Re
                 Service::getVersionDependentDatabaseColumnName('o_key'));
         }
 
+        if ($isTagsSet) {
+            $tags = is_array($args['tags']) ? $args['tags'] : explode(',', $args['tags']);
+            if ($tags) {
+                foreach ($tags as $tagName) {
+                    $conditionParts[] = "id IN (
+                        SELECT cId FROM tags_assignment INNER JOIN tags ON tags.id = tags_assignment.tagid
+                    WHERE ctype = 'object' AND name = " . $tagName . ")";
```
NiklasBr, line 307:
```suggestion
                    $conditionParts[] = 'id IN (
                        SELECT cId FROM tags_assignment INNER JOIN tags ON tags.id = tags_assignment.tagid
                    WHERE ctype = 'object' AND name = ' . Db::get()->quote($tagName) . ')';
```

Issue #657, opened 2022-12-29
[Bug]: Some Operators don't work


robertSt7:
### Expected behavior

Operators don't throw an exception

### Actual behavior

Operators (Substring, ... ) throw an exception

### Steps to reproduce

- Add the Substring-Operator to an Query Schema
- Query the Operator

PR #671, opened 2023-01-20
Deprecate merge operator

lukmzig:
Mark the merge operator as deprecated since it is not used anymore. 
Related to #657


mcop1:
I noticed that we don´t use translated values for the name of the operator. Created a follow up for that: 
https://github.com/pimcore/data-hub/issues/674

PR #678, opened 2023-01-24
[Task] Add pimcore 10 and 11 support for new menu creation

### LINKED ISSUE #677: [Improvement]: Adjust navigation creation
### Improvement description

Improvement description
Creating own datahub class instead of just adding the buttons via event listeners. Use the new approach based on the preMenuBuild Event

Based on https://github.com/pimcore/pimcore/pull/14113

mattamon:
## Changes in this pull request  
Resolves #677 

## Additional info  
If new preMenuBuild Event is available use the new methods, if not use the pimcoreReady Event to setup the menu

### DIFF
--- a/src/Resources/public/css/icons.css
+++ b/src/Resources/public/css/icons.css
@@ -1,3 +1,7 @@
+.pimcore_main_nav_icon_mind_map {
+    background: url(/bundles/pimcoreadmin/img/flat-white-icons/mind_map.svg) center center no-repeat;
+}
+
 .plugin_pimcore_datahub_icon  {
     background: url(/bundles/pimcoreadmin/img/twemoji/2728.svg) center center no-repeat !important;
 }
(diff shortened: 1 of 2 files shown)


Corepex:
Could you please add a permission check before adding the menu item to the toolbar?

https://github.com/pimcore/data-hub/blob/5afaef0611df148bdfbc1d9902f9b1aaa94e21df/src/Resources/public/js/datahub.js#L53-L63

I noticed that the toolbar icon is added, although the user doesn't have the permission to open it.  
This was checked in the previous version ....

https://github.com/pimcore/data-hub/blob/416a77c790b6510fd628b33b0be122520dc7bf3d/src/Resources/public/js/datahub.js#L21-L37

How to reproduce:
create a user and do not add `DataHub Admin` or `DataHub Config` permissions.

mattamon:
Works for Pimcore 10 and 11 now. 
Main difference are the events pimcoreReady and preMenuBuild. 

If we ever kick the pimcoreReady event we can remove the eventlistener completely.

PR #694, opened 2023-02-09
[Bug]: Can't fetch data of editables for documents with inheritance

### LINKED ISSUE #570: [Bug]: Can't fetch data of editables for documents with inheritance
### Expected behavior

I expect editables and their content on documents or headless documents that have been pasted via inheritance (with Content-Master Document) to be exposed via GraphQL.

image

### Actual behavior

**Documents or headless documents that have been copied and pasted using inheritance (setting a Content-Master Document) do not expose their content and editables via GraphQL.** 

Picture below is showing query and response for:
1.  "original" document and its editables, 
2. pasted document not using inheritance (see "paste") showing same editables and 
3. pasted document using inheritance (see "paste_inheritance") showing empty field editables

image

Remark: Content and editables are getting exposed only when they are overwritten in copied document with inheritance.

### Steps to reproduce

In Pimcore Demo …

- Copy a document e.g. "Data Quality" (ID: 193)
- Paste this document using inheritance (of structure and content)
- Try to query for pasted document and its editables via GraphQL
- Field **editables is empty**

mcop1:
Resolves #570

Added editables for ContentMasterDocument

### DIFF
--- a/src/GraphQL/DocumentType/PageSnippetType.php
+++ b/src/GraphQL/DocumentType/PageSnippetType.php
@@ -57,7 +57,7 @@ class PageSnippetType extends AbstractDocumentType
                 'getInheritedValues' => [
                     'type' => Type::boolean(),
                     'description' => 'Whether inherited editables should be fetched or not.',
-                    'defaultValue' => [false]
+                    'defaultValue' => false
                 ],
             ],
             'resolve' => [$resolver, 'resolveElements']
--- a/src/GraphQL/DocumentType/PageSnippetType.php
+++ b/src/GraphQL/DocumentType/PageSnippetType.php
@@ -53,6 +53,13 @@ class PageSnippetType extends AbstractDocumentType
 
         $config['fields']['editables'] = [
             'type' => Type::listOf($this->documentElementType),
+            'args' => [
+                'getInheritedValues' => [
+                    'type' => Type::boolean(),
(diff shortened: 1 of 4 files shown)


blankse:
Why not bump pimcore version in composer.json to 10.5.16?

mcop1:
> Why not bump pimcore version in composer.json to 10.5.16?

Was thinking about that too, but I didn´t want to change the requirement in a bugfix release.
But seeing the problems it would be the easier way.

blankse:
But is already 10.5, so you bump only bugfix releases. Think this is fine, isn't it?

mcop1:
> But is already 10.5, so you bump only bugfix releases. Think this is fine, isn't it?

Agreed, should be fine 👍

src/GraphQL/DocumentType/PageSnippetType.php
```diff
@@ -53,6 +53,13 @@ public function build(&$config)
 
         $config['fields']['editables'] = [
             'type' => Type::listOf($this->documentElementType),
+            'args' => [
+                'getInheritedValues' => [
+                    'type' => Type::boolean(),
+                    'description' => 'Whether inherited editables should be fetched or not.',
+                    'defaultValue' => [false]
```
robertSt7:
```suggestion
                    'defaultValue' => false
```

PR #702, opened 2023-02-28
Add tests with php 8.2

blankse:


blankse:
The tests a failing because of a BC break in 10.6. See https://github.com/pimcore/pimcore/pull/14263

Now it is used the symfony-config instead of the settings-store (although the env variable is set)

blankse:
@mcop1 The tests pass now.

mcop1:
Thank you :)

PR #703, opened 2023-03-01
Fix error with localizedFields in object bricks

blankse:
image

### DIFF
--- a/src/GraphQL/FieldHelper/DataObjectFieldHelper.php
+++ b/src/GraphQL/FieldHelper/DataObjectFieldHelper.php
@@ -172,22 +172,20 @@ class DataObjectFieldHelper extends AbstractFieldHelper
                 $fieldDefinitionLocalizedFields = $brickDefinition->getFieldDefinition('localizedfields');
                 if ($fieldDefinition = $fieldDefinitionLocalizedFields?->getFieldDefinition($brickKey)) {
                     $container = $fieldDefinitionLocalizedFields;
-                } else {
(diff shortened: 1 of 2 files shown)


blankse:
It was also not possible to set the language argument:
`Unknown argument \"language\" on field \"xyz\" of type \"object_Xyz\".`

src/GraphQL/FieldHelper/DataObjectFieldHelper.php
```diff
@@ -168,12 +168,12 @@ public function getFieldDefinitionFromKey($class, $key, &$container = null)
             $brickDefinition = Definition::getByKey($brickType);
 
             if ($brickDescriptor) {
-                $fieldDefinition = $brickDefinition->getFieldDefinition($brickKey);
```
blankse, line 171: It seems the getFieldDefinition() method of the object brick definition also get localized field definitions. So the $container variable isn't set for localized fields.

blankse:
I changed it also for the class field definitions because the getFieldDefinition method in 11.x loads also the localized fields: https://github.com/pimcore/pimcore/pull/14105

src/GraphQL/FieldHelper/DataObjectFieldHelper.php
mcop1: You are right, that´s why we unified the behavior for pimcore 11 (see your other comment). Thanks!

mcop1:
As always, thank you 👍

PR #646, opened 2022-12-12
[Improvement] Replace childs with children

robertSt7:
resolves #607

PR #706, opened 2023-03-06
Fix migration Version20221212152145

blankse:
Followup to https://github.com/pimcore/data-hub/pull/646
I got following error:
```
SQLSTATE[42000]: Syntax error or access violation: 1142 UPDATE command denied to user 'db'@'172.31.0.3' for table `pimcore`.`settings_store`
```
Not every database is named 'pimcore'


mcop1:
Thanks!

Edit: Will merge it as soon as we have fixed the bc break

mcop1:
Tests are working now, thank you again :)

PR #708, opened 2023-03-15
Prevent undefined array key errors on Link editables

DuckThom:
When a Link editable is added to a template, an `Undefined array key '<field>'` error is returned when a data field queried and the editable is not configured. In those cases, `$value->getData()` returns an array but without certain fields such as `title` or the `internal` fields


blankse:
@DuckThom Looks like a bugfix to me. Can you rebase it on the 1.5 branch?

mcop1:
Thank you

PR #711, opened 2023-03-20
Update webonyx/graphql-php to v15

rubanooo:
I have updated webonyx/graphql-php to the latest version and applied the needed PHPStan updates

### DIFF
--- a/src/GraphQL/General/UploadType.php
+++ b/src/GraphQL/General/UploadType.php
@@ -27,7 +27,7 @@ class UploadType extends ScalarType
 {
     public string $name = 'Upload';
 
-    public null|string $description =
+    public null | string $description =
         'The `Upload` special type represents a file to be uploaded in the same HTTP request as specified by
  [graphql-multipart-request-spec](https://github.com/jaydenseric/graphql-multipart-request-spec).';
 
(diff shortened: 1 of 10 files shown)


src/GraphQL/General/UploadType.php
```diff
@@ -25,15 +25,9 @@
...
-     * @var string
-     */
-    public $name = 'Upload';
+    public string $name = 'Upload';
 
-    /**
-     * @var string
-     */
-    public $description =
+    public null | string $description =
```
mcop1:
```suggestion
    public string $description =
```

Just a minor thing, but there is no reason for the description to allow null values, right?

rubanooo: It has to be defined `null | string` because it extends the base type class, PHPStan will throw an error when you don't

mcop1:
```suggestion
    public ?string $description =
```
Aaah you are right, of course. Sorry, I missed that. Maybe we can change it to `?string` though, since this is the style we want to use from Pimcore 11 on.

mcop1:
@rubanooo Could you please merge the data-hub 1.x branch in to your pr-branch again? This should resolve the failing tests. 
Thank you for your help.

mcop1:
Thanks again :)

PR #714, opened 2023-03-24
Adapt Admin Bundle refactorings

dvesh3:
Related to https://github.com/pimcore/pimcore/pull/14642


mcop1:
Regarding tests, please see follow up ticket https://github.com/pimcore/data-hub/issues/721

PR #717, opened 2023-03-29
[Improvement]: Add configuration storage locations to Symfony tree

### LINKED ISSUE #710: [Improvement]: Add configuration storage locations to Symfony tree
### Improvement description

see https://github.com/pimcore/pimcore/issues/12066

robertSt7:
resolves #710

### DIFF
--- a/src/Configuration/Dao.php
+++ b/src/Configuration/Dao.php
@@ -46,19 +46,32 @@ class Dao extends Model\Dao\PimcoreLocationAwareConfigDao
      */
     private const LEGACY_FILE = 'datahub-configurations.php';
 
+    /**
+     * @deprecated Will be removed in Pimcore 11
+     */
     public const CONFIG_PATH = PIMCORE_CONFIGURATION_DIRECTORY . '/data-hub';
 
     public function configure(): void
     {
         $config = \Pimcore::getContainer()->getParameter('pimcore_data_hub');
 
-        parent::configure([
-            'containerConfig' => $config['configurations'] ?? [],
-            'settingsStoreScope' => 'pimcore_data_hub',
-            'storageDirectory' => self::CONFIG_PATH,
-            'legacyConfigFile' => self::LEGACY_FILE,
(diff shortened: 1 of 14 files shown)


src/Configuration/Dao.php
```diff
@@ -46,19 +46,32 @@ class Dao extends Model\Dao\PimcoreLocationAwareConfigDao
      */
     private const LEGACY_FILE = 'datahub-configurations.php';
 
+    /**
+     * @deprecated Will be removed in Pimcore 11
```
mcop1: It will be removed as soon as we drop 10.6 support, right? So it is not 100% correct. Could you please check other deprecations too? 

```suggestion
     * @deprecated Will be removed as soon as Pimcore 10.6 isn´t supported anymore.
```

src/GraphQL/Query/Operator/Thumbnail.php
```diff
@@ -64,7 +64,7 @@ public function getLabeledValue($element, ResolveInfo $resolveInfo = null)
             if ($childResult->value instanceof Asset\Image || $childResult->value instanceof Asset\Video) {
                 $childValue = $result->value = $childResult->value;
                 $thumbnail = $childValue->getThumbnail($this->thumbnailConfig, false);
-                $result->value = $thumbnail->getPath(false);
```
mcop1: That change shouldn´t be in this pr

mcop1 (commented):
We should add documentation for that too, maybe somewhere here (separate sections for Pimcore 10.6 and 11)
https://github.com/pimcore/data-hub/blob/54673a55ca691a3fe3bf152c59f12e067e890e6d/doc/20_Deployment.md#L31

mcop1:
One more thing. What about tests? We are using the env variable here, but that will not work with 11 anymore, right?
https://github.com/pimcore/data-hub/blob/9139c1478cc798b5264f2941435ce4f7f6b6c06c/.github/ci/files/.env.test#L1

mcop1:
Regarding tests, please see the follow up https://github.com/pimcore/data-hub/issues/721

PR #715, opened 2023-03-28
[Task] Correct POEditor workflow and fix compatability to Pimcore 11

Corepex:
follow-up for: https://github.com/pimcore/data-hub/pull/713

PR #718, opened 2023-03-30
Fix: TypeErrors

robertSt7:


mcop1:
Same as here, right? https://github.com/pimcore/data-hub/pull/715/files 
Would prefer the solution in this pr.

PR #739, opened 2023-05-03
[Task]: Add childs to children migration

### LINKED ISSUE #737: [Task]: Add `childs` to `children` migration
### Improvement description

`childs` nodes should be renamed to `children`. Migration for settings-store is already existing, but symfony config is missing. 

Follow up to https://github.com/pimcore/data-hub/pull/646

robertSt7:
resolves #737

### DIFF
--- a/src/Migrations/PimcoreX/Version20230503165847.php
+++ b/src/Migrations/PimcoreX/Version20230503165847.php
@@ -41,6 +41,7 @@ final class Version20230503165847 extends AbstractMigration
 
     private function replaceStringInFiles(string $search, string $replace): void
     {
+    if(is_dir(self::CONFIG_DIR)){
         $files = scandir(self::CONFIG_DIR, 0);
         for ($i = 2; $i < count($files); $i++) {
             $file = self::CONFIG_DIR . '/' . $files[$i];
@@ -48,5 +49,6 @@ final class Version20230503165847 extends AbstractMigration
             $fileContent = str_replace($search, $replace, $fileContent);
             file_put_contents($file, $fileContent);
         }
+    }
     }
 }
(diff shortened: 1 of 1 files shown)


mcop1 (commented):
@robertSt7 What happens if someone overwrites the storage location with the corresponding env variable?

robertSt7:
> @robertSt7 What happens if someone overwrites the storage location with the corresponding env variable?

That was not possible for Data-Hub until the BC-Layer in 10.6

src/Migrations/PimcoreX/Version20230503165847.php
```diff
@@ -0,0 +1,52 @@
...
+
+    private function replaceStringInFiles(string $search, string $replace): void
+    {
+        $files = scandir(self::CONFIG_DIR, 0);
+        for ($i = 2; $i < count($files); $i++) {
+            $file = self::CONFIG_DIR . '/' . $files[$i];
+            $fileContent = file_get_contents($file);
+            $fileContent = str_replace($search, $replace, $fileContent);
+            file_put_contents($file, $fileContent);
+        }
```
Corepex:
```suggestion
    if(is_dir(self::CONFIG_DIR)){
        $files = scandir(self::CONFIG_DIR, 0);
        for ($i = 2; $i < count($files); $i++) {
            $file = self::CONFIG_DIR . '/' . $files[$i];
            $fileContent = file_get_contents($file);
            $fileContent = str_replace($search, $replace, $fileContent);
            file_put_contents($file, $fileContent);
        }
    }
```

since `scandir` will throw an error if the directory issnt there I would wrap it in a `is_dir` condition.

podarcis:
@robertSt7 Before I open a new report I'd like to clarify: As this migration is also run in 1.6, which lists Pimcore 10.6 as possible dependency, isn't this migration too early as Pimcore 10.6 doesn't have "childs" replaced all over (e.g. Grid Operators)? I couldn't properly load the config in the UI when this migration happened to yaml files and needed to revert them.

PR #768, opened 2023-07-13
fix: include YouTube/Vimeo id in DocumentElementType\VideoType

youwe-petervanderwal:
Error: The field was included, but didn't contain a value
Resolved with: Adding a resolve function and changing type to string (as a YouTube id contains non-numeric characters)

### DIFF
--- a/src/GraphQL/DocumentElementType/VideoType.php
+++ b/src/GraphQL/DocumentElementType/VideoType.php
@@ -58,7 +58,12 @@ class VideoType extends ObjectType
                             }
                         ],
                         'id' => [
-                            'type' => Type::int()
+                            'type' => Type::string(),
+                            'resolve' => static function ($value = null, $args = [], $context = [], ResolveInfo $resolveInfo = null) {
+                                if ($value instanceof Video) {
+                                    return $value->getId();
+                                }
+                            }
                         ],
                         'type' => [
                             'type' => Type::string(),


robertSt7:
HI @youwe-petervanderwal  thanks for reporting this bug. We are not sure if this is a BC-Break when an implementation relies  on the return type int. Could you please test this? Thank you

youwe-petervanderwal:
> We are not sure if this is a BC-Break when an implementation relies on the return type int. Could you please test this?

I do understand this concern. During testing I found the following

a) Although the property was typed `Int`, it **always** returned `null` as it didn't had a resolver configured. So, including the property in your QraphQL query was useless.

b) There are two editable types with a property `id`. The other one (`document_editableRelation`) contains the Pimcore element id and is therefore also marked as `Int`. The following query was (before my change) valid but useless (see remark `a`) and will now throw an error due to mixed types.

```
query {
  getDocument(id: 29){
    ...on document_page {
      editables {
        __typename
        ...on document_editableVideo {
          id
        }
        ...on document_editableRelation {
          id
        }
      }
    }
  }
}
```

That can be resolved though with providing an alias

```
query {
  getDocument(id: 29){
    ...on document_page {
      editables {
        __typename
        ...on document_editableVideo {
          videoId: id
        }
        ...on document_editableRelation {
          id
        }
      }
    }
  }
}
```

As there are more editables with conflicting properties (e.g. `data` on `document_editableAreablock`, `document_editableLink` and `document_editableScheduledblock`) I don't think having this name conflict a problem on its own.

If we don't want this 'BC-Break' (although I don't think it would affect much applications as it was a useless property), I can add it as a new property (e.g. `videoId`) now and create a separate PR for a next major version to remove the useless `id` property.

Personally I do think the name `id` makes more sense as this is also the name in the `\Pimcore\Model\Document\Editable\Video` class.

robertSt7:
> > We are not sure if this is a BC-Break when an implementation relies on the return type int. Could you please test this?
> 
> I do understand this concern. During testing I found the following
> 
> a) Although the property was typed `Int`, it **always** returned `null` as it didn't had a resolver configured. So, including the property in your QraphQL query was useless.
> 
> b) There are two editable types with a property `id`. The other one (`document_editableRelation`) contains the Pimcore element id and is therefore also marked as `Int`. The following query was (before my change) valid but useless (see remark `a`) and will now throw an error due to mixed types.
> 
> ```
> query {
>   getDocument(id: 29){
>     ...on document_page {
>       editables {
>         __typename
>         ...on document_editableVideo {
>           id
>         }
>         ...on document_editableRelation {
>           id
>         }
>       }
>     }
>   }
> }
> ```
> 
> That can be resolved though with providing an alias
> 
> ```
> query {
>   getDocument(id: 29){
>     ...on document_page {
>       editables {
>         __typename
>         ...on document_editableVideo {
>           videoId: id
>         }
>         ...on document_editableRelation {
>           id
>         }
>       }
>     }
>   }
> }
> ```
> 
> As there are more editables with conflicting properties (e.g. `data` on `document_editableAreablock`, `document_editableLink` and `document_editableScheduledblock`) I don't think having this name conflict a problem on its own.
> 
> If we don't want this 'BC-Break' (although I don't think it would affect much applications as it was a useless property), I can add it as a new property (e.g. `videoId`) now and create a separate PR for a next major version to remove the useless `id` property.
> 
> Personally I do think the name `id` makes more sense as this is also the name in the `\Pimcore\Model\Document\Editable\Video` class.

@youwe-petervanderwal You are right, we should change the return type and add a resolver. Thanks for finding the bug and creating the PR :)

PR #773, opened 2023-07-25
WebserviceController: query must be a string

blankse:
Fix https://github.com/pimcore/data-hub/issues/772


dvesh3:
@blankse thanks!

PR #784, opened 2023-08-07
[Bug]: Pimcore 10: Symfony config has no effect on the storage type

### LINKED ISSUE #779: [Bug]: Pimcore 10: Symfony config has no effect on the storage type
### Expected behavior

Symfony config is used for reading and writing data hub config files

### Actual behavior

Symfony config has no effect

### Steps to reproduce

Pimcore 10.6

1. Add config:
```yaml
pimcore_data_hub:
    config_location: 
        data_hub:
            write_target:
                type: 'symfony-config'
                options:
                     directory: '/var/www/html/var/config/data-hub-custom'
```

2. Add new datahub graphql configuration
3. Config is still stored on default storage in db table `settings-store`

Reason: the config is used only on v11 https://github.com/pimcore/data-hub/blob/323590cf0952d0f958e8b3f15757f7a7cd679f64/src/DependencyInjection/PimcoreDataHubExtension.php#L57-L60

robertSt7:
resolves #779

### DIFF
--- a/src/DependencyInjection/PimcoreDataHubExtension.php
+++ b/src/DependencyInjection/PimcoreDataHubExtension.php
@@ -16,6 +16,7 @@
 namespace Pimcore\Bundle\DataHubBundle\DependencyInjection;
 
 use Pimcore\Bundle\CoreBundle\DependencyInjection\ConfigurationHelper;
+use Pimcore\Bundle\DataHubBundle\Configuration\Dao;
 use Symfony\Component\Config\FileLocator;
 use Symfony\Component\DependencyInjection\ContainerBuilder;
 use Symfony\Component\DependencyInjection\Extension\Extension;
@@ -60,14 +61,13 @@ class PimcoreDataHubExtension extends Extension implements PrependExtensionInter
             new FileLocator([$configDir, Dao::CONFIG_PATH])
         );
 
-        //load datahub configs
-        $configs = ConfigurationHelper::getSymfonyConfigFiles($configDir);
 
         //TODO: remove as soon as Pimcore 10.6 isn´t supported anymore.
         $configLocator = new \Pimcore\Bundle\DataHubBundle\Configuration\DatahubConfigLocator();
         $configs =
             [
-                ...$configs,
+                ...ConfigurationHelper::getSymfonyConfigFiles($configDir),
+                ...ConfigurationHelper::getSymfonyConfigFiles($_SERVER['PIMCORE_CONFIG_STORAGE_DIR_DATA_HUB'] ?? ''),
                 ...$configLocator->locate('config')
             ];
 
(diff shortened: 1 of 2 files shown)


robertSt7:
Waiting for pimcore 10.6.8 release

dvesh3:
@robertSt7 why do we need `10.6.8` for this?

robertSt7:
> @robertSt7 why do we need `10.6.8` for this?

@dvesh3 because we need this https://github.com/pimcore/pimcore/pull/15695

aryaantony92 (approved):

src/DependencyInjection/PimcoreDataHubExtension.php
```diff
@@ -52,29 +52,18 @@ public function prepend(ContainerBuilder $container)
...
-            $configs = ConfigurationHelper::getSymfonyConfigFiles($configDir);
-            foreach ($configs as $config) {
-                $configLoader->load($config);
-            }
-        } else {
-            $configLocator = new \Pimcore\Bundle\DataHubBundle\Configuration\DatahubConfigLocator();
-            foreach ($configLocator->locate('config') as $config) {
-                $configLoader->load($config);
-            }
+        $configs = ConfigurationHelper::getSymfonyConfigFiles($configDir);
```
dvesh3: so now this only considers the directory from the Symfony config not the default [dir ](https://github.com/pimcore/data-hub/pull/784/files#diff-d3cd915a931a9be0a3278d79a468a433bd6016436b5a82f9188e72181d3170c5L55) anymore. what happens now if i use `PIMCORE_CONFIGURATION_DIRECTORY`?

robertSt7: @dvesh3 I didn't find any docs with `PIMCORE_CONFIGURATION_DIRECTORY` should we consider it anyway here?

aryaantony92: @dvesh3 Did you mean suppose if we didn't give any write_target in the yaml configuration, what will be the value of $configDir ? In that case, it will be set to '/var/config/data_hub'. https://github.com/pimcore/data-hub/blob/e28c7c4523367efa768e34aa86cb2266cc1468c0/src/DependencyInjection/Configuration.php#L52

dvesh3: I mean we don't use [`DatahubConfigLocator`](https://github.com/pimcore/data-hub/blob/30ac047f8ace8a6fbfc5ed156bdc3348d28a38a0/src/Configuration/DatahubConfigLocator.php#L42) anymore which uses https://github.com/pimcore/data-hub/blob/e7c97fcaa62446e7a1fae5f3a280728cb67f8fa2/src/Configuration/Dao.php#L52. that means if i change the constant to some custom location this will not work. right?

robertSt7: That's right, so I will fix this. Thanks for finding this

PR #802, opened 2023-09-21
Expand on tag filters

abrenoch:
I noticed the `tags` filter was a bit rigid so I have some proposed changes:

- Adds `useTagIdPath` as a query variable. This will allow searching for tags by their nested location (such as `/7/11/15/`) rather than just the name. If 2 tags with the same name existed within 2 different parents, there was no way to differentiate which one was returned. This will allow targeting specific tags within the scope of the parents.
- Adds `hasAllTags` as a query variable. Currently the tag filter matches *any* of the tags provided in the query, setting this variable to `true` will require results match *all* the tags provided.


abrenoch:
I'm going to re-open this - the commits must not have my email and the CLAassistant isn't recognizing my signature

Issue #804, opened 2023-09-25
[Improvement]: Use 'getThumbnail()' instead of 'getThumbnailConfig()'


aryaantony92:
### Improvement description

Related to https://github.com/pimcore/pimcore/pull/15869

- [x] https://github.com/pimcore/data-hub/blob/ce780dd420818e49914da373fb869f8865b5060c/src/GraphQL/Resolver/AssetType.php#L196

blankse:
@aryaantony92 I created a PR https://github.com/pimcore/data-hub/pull/805

aryaantony92:
@blankse Thanks!

aryaantony92:
Fixed by https://github.com/pimcore/data-hub/pull/805

PR #805, opened 2023-09-25
Use 'getThumbnail()' instead of 'getThumbnailConfig()'

blankse:
Resolve https://github.com/pimcore/data-hub/issues/804


aryaantony92:
@blankse Thanks again

PR #812, opened 2023-11-05
Add mutation datatype 'input quantity value'

leitge:
Solves issue:
Mutation datatype 'quantityValue' not supported yet


Corepex:
@leitge thanks for your contribution :tada: :man_astronaut:

PR #815, opened 2023-11-08
Sort items in the datahub

jeremygnoonan:
Sorts data hub list:

~~* Folders come first, sorted~~
* Items in folders are sorted
* Things outside of folders are sorted and included second

### DIFF
--- a/src/Controller/ConfigController.php
+++ b/src/Controller/ConfigController.php
@@ -115,19 +115,17 @@ class ConfigController extends \Pimcore\Controller\UserAwareController
             }
         }
 
+        $sortFunc = fn($a, $b)  =>  strtolower($a['text']) <=> strtolower($b['text']);
+
         //sort group children
         foreach($groups as &$group){
-            usort($group['children'], function ($a, $b) {
-                return $a['text'] <=> $b['text'];
-            });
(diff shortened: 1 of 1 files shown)


NiklasBr:
Please don't sort folders first, it is unintuitive and breaks expectations.

jeremygnoonan:
@NiklasBr I removed the folder first sorting, and also made it case insensitive to behave like the object tree.

mcop1:
Hello @jeremygnoonan,
thank you for the pr, code works fine. 

What do you think about sorting by type too? With different data-hub plugins it would look something like this right now: 
image

Personally I would like it a bit more if we sort by plugin type too. But I guess that would go in the same directions as sorting the folders. Any opinions on that? @NiklasBr ?

mcop1:
To keep things short, I created a follow up ticket to sort configurations in the front end (where sorting should really be imo). Feedback is welcome of course -> https://github.com/pimcore/data-hub/issues/826 

Thanks again for the pr!

NiklasBr:
Just alphabetical natural sorting IMO is great, everything else is known to cause undue cognitive load.

jeremygnoonan:
Thanks guys hope you had a great holiday!

PR #816, opened 2023-11-10
[Bug]: GraphQl - query for asset meta data returns an exception

### LINKED ISSUE #811: [Bug]: GraphQl - query for asset meta data returns an exception
### Expected behavior

List of asset meta data is returned

### Actual behavior

Exception is thrown ( "Expected a value of type String but received: [[\"object\",12]]. String cannot represent value: [[\"object\",12]]")

### Steps to reproduce

1. Go to https://demo.pimcore.com/
2. Open graphQl **assets** configuration and add _Asset Listing_ read permission in the Schema Definition and save configuration
4. Open playground or go to https://demo.pimcore.com/pimcore-datahub-webservices/explorer/assets?apikey=9e265983fbd2dc42b8f7cdb841c82807
5. Add Query to list asset meta data with data e.g
`{
  getAssetListing {
    totalCount
    edges {
      node {
       ... on asset { 
      		id
        	fullpath
        metadata(ignore_language: true) { name type data}
      	}
      }
    }
  }
}`
6. Exception is returned (with debugger exception states:  "Expected a value of type String but received: [[\"object\",12]]. String cannot represent value: [[\"object\",12]]")

robertSt7:
resolves #811

### DIFF
--- a/src/GraphQL/Resolver/AssetType.php
+++ b/src/GraphQL/Resolver/AssetType.php
@@ -17,11 +17,14 @@ namespace Pimcore\Bundle\DataHubBundle\GraphQL\Resolver;
 
 use Exception;
 use GraphQL\Type\Definition\ResolveInfo;
+use Pimcore\Bundle\DataHubBundle\Event\GraphQL\AssetMetadataEvents;
 use Pimcore\Bundle\DataHubBundle\GraphQL\ElementDescriptor;
 use Pimcore\Bundle\DataHubBundle\GraphQL\Traits\ElementTagTrait;
 use Pimcore\Bundle\DataHubBundle\GraphQL\Traits\ServiceTrait;
 use Pimcore\Bundle\DataHubBundle\WorkspaceHelper;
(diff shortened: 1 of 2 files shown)


src/GraphQL/Resolver/AssetType.php
```diff
@@ -64,39 +67,48 @@ public function resolveTag($value = null, $args = [], $context = [], ResolveInfo
...
+            return null;
+        }
 
-        if ($asset) {
-            $metadata = $asset->getObjectVar('metadata');
-            if ($metadata) {
-                if (isset($args['ignore_language']) && $args['ignore_language']) {
-                    return $metadata;
-                }
+        //"object_154,asset_489"
```
kingjia90: ahh, this is to hint the format returned from the event, right? can we wrap it with some description? 
something like
```suggestion
        //Relational data would return as a comma separated list. each element type and id by `_`, eg. "object_154,asset_489"
```

PR #825, opened 2023-12-14
[Bug][Data loss]: if passing an invalid asset.id, fullpath to object mutation GraphQL query, the assets are silently discarded

### LINKED ISSUE #823: [Bug][Data loss]: if passing an invalid asset.id, fullpath to object mutation GraphQL query, the assets are silently discarded
### Expected behavior

Error reported before any mutation is done. Same thing happens with unknown IDs, it just ignores them.

### Actual behavior

Any unknown assets silently discarded.

### Steps to reproduce

1. create class `Product` with input `SKU` and many to one assets called `Images`
2. add an object 123 with a few images
3. run this (see #822 why the path contains a hostname):
```gql
mutation {
  updateProduct(
    id: 123
    input: {
      SKU: "ABC"
      Images: [
        {
          fullpath: "https://example.com/Brands/loxeal_logo.png"
          type: "asset"
        }
      ]
    }
  ) {
    output {
      id
    }
    message
    success
  }
}
```
4. it responds with success and the SKU is correctly persisted, but images are empty (and any previous images are deleted)

robertSt7:
resolves #823

### DIFF
--- a/src/GraphQL/DataObjectInputProcessor/ManyToManyRelation.php
+++ b/src/GraphQL/DataObjectInputProcessor/ManyToManyRelation.php
@@ -48,7 +48,7 @@ class ManyToManyRelation extends Base
                         $result[] = $element;
                     } else {
                         throw new NotFoundException(
-                            "Doesn't found element: " . $newValueItemValue['fullpath'] ?? $newValueItemValue['id']
+                            "Doesn't found element: " . $newValueItemValue['fullpath']
                         );
                     }
                 }
--- a/src/GraphQL/DataObjectInputProcessor/ManyToManyRelation.php
+++ b/src/GraphQL/DataObjectInputProcessor/ManyToManyRelation.php
@@ -20,6 +20,7 @@ use Pimcore\Bundle\DataHubBundle\GraphQL\Service;
 use Pimcore\Bundle\DataHubBundle\GraphQL\Traits\ElementIdentificationTrait;
 use Pimcore\Model\DataObject\Concrete;
 use Pimcore\Model\DataObject\Fieldcollection\Data\AbstractData;
+use Pimcore\Model\Exception\NotFoundException;
 
 class ManyToManyRelation extends Base
 {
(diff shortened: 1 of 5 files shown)


src/GraphQL/DataObjectInputProcessor/ManyToManyRelation.php
```diff
@@ -45,6 +46,10 @@ public function process($object, $newValue, $args, $context, ResolveInfo $info)
 
                     if ($element) {
                         $result[] = $element;
+                    } else {
+                        throw new NotFoundException(
+                            "Doesn't found element: " . $newValueItemValue['fullpath']
```
kingjia90: Looks like grammatically incorrect and only covering `fullpath`, should be more like `Element with id or path not found`, see also [getElementByTypeAndIdOrPath](https://github.com/pimcore/data-hub/blob/3a454f2ac097e3d4b2e99a5eba92e4e087d72b61/src/GraphQL/Traits/ElementIdentificationTrait.php#L39-L65) on possible errors, since it is what returned to `$element` on https://github.com/pimcore/data-hub/blob/b76286cc0ade02e027eafe4a9e33c9e735ada7bf/src/GraphQL/DataObjectInputProcessor/ManyToManyRelation.php#L45

kingjia90:
Also extend the change to all the other relational types eg.

https://github.com/pimcore/data-hub/blob/630d7295fca25310f7f5ea897d67e36c9761d5c4/src/GraphQL/DataObjectInputProcessor/AdvancedManyToManyRelation.php#L47-L60
https://github.com/pimcore/data-hub/blob/3a454f2ac097e3d4b2e99a5eba92e4e087d72b61/src/GraphQL/DataObjectInputProcessor/ManyToManyObjectRelation.php#L51-L53

or maybe just move the exception to `getElementByTypeAndIdOrPath()` but may need to change these 
https://github.com/pimcore/data-hub/blob/e28c7c4523367efa768e34aa86cb2266cc1468c0/src/GraphQL/Resolver/QueryType.php#L214-L217
to `try-catch`

PR #833, opened 2024-02-05
[Bug]: 'o_id' column not found error when use tags args for listing queries

### LINKED ISSUE #834: [Bug]: 'o_id' column not found error when use tags args for listing queries
### Expected behavior

Get List of Data Objects selected according to the specified tags (as argument) by GraphQL query

for ex:
изображение

### Actual behavior

Getting error: "An exception occurred while executing a query: SQLSTATE[42S22]: Column not found: 1054 Unknown column 'o_id' in 'IN/ALL/ANY subquery'"

изображение

### Steps to reproduce

When using pimcore > v.11:
Query list of object with "tags" arg

bender-nvkz:


bender-nvkz:
Reported as bug in #834

fashxp:
thx very much.

PR #837, opened 2024-02-24
Add QueryType for Version on Objects

leitge:


src/GraphQL/DataObjectType/PimcoreObjectType.php
```diff
@@ -83,6 +83,20 @@ public function build($context = [])
...
             'modificationDate' => Type::int(),
+            'version' => [
+                'type' => Type::int(),
+                'resolve' => function ($value = null, $args = [], $context = [], ResolveInfo $resolveInfo = null) {
+                    $object = \Pimcore\Model\DataObject::getById($value['id']);
+                    if ($object) {
+                        $version = $object->getVersions()[count($object->getVersions()) - 1]->getId();
+                        if ($version) {
+                            return $version;
+                        }
```
blankse: I think there could be an error if you delete all versions.
```suggestion
                        $versions = $object->getVersions();
                        if ($versions) {
							$version = $versions[count($versions) - 1]->getId();
							if ($version) {
								return $version;
							}
                        }
```

leitge: @blankse how do you mean that? Even if I delete all versions of an object, one published version always remains. The GraphQL query continues to work locally, at least for me, and outputs the last published version

blankse: I also think the last version is not enough. It could be unpublished draft versions. We need to compare the modification date:
```suggestion
                        foreach ($object->getVersions() as $version) {
							if ($object->getModificationDate() === $version->getDate()) {
								return $version->getId();
							}
                        }
```

leitge: @blankse agree to both topics. 

Even when I initially thought the 2nd one wasn't necessary. It is necessary! To get to the result faster in the loop (if people save countless versions) I would run the array backwards. 

Thanks for your support 💪🫡


```suggestion
                        foreach (array_reverse($object->getVersions()) as $version) {
                            if ($object->getModificationDate() === $version->getDate()) {
                                return $version->getId();
                            }
                        }
```

Corepex:
@leitge, thanks for your PR. Tested your code and it just works - great work :1st_place_medal: :man_astronaut:

PR #840, opened 2024-02-29
PHP 8.3 Support

blankse:
See https://github.com/pimcore/pimcore/pull/16589

Requires https://github.com/pimcore/compatibility-bridge-v10/pull/6


Corepex (approved):
LGTM

fashxp:
thank you very much. 
merging that int the bugfix branch since next minor is just half the way finished and we don't want to wait for it.

PR #854, opened 2024-04-22
Upgraded to PHP CS Fixer v3

blankse:
See other pimcore repositories


.github/workflows/php-cs-fixer.yaml
```diff
@@ -1,19 +1,26 @@
...
         steps:
             - uses: actions/checkout@v2
               with:
                   ref: ${{ github.head_ref }}
 
             - name: PHP-CS-Fixer
-              uses: docker://oskarstark/php-cs-fixer-ga:2.19.0
-              with:
-                args: --config=.php_cs.dist --allow-risky yes
+              uses: docker://oskarstark/php-cs-fixer-ga:latest
```
fashxp: should we really go for the latest, or tag at least for a major version?

fashxp (changes requested):

.github/workflows/php-cs-fixer.yaml
blankse: See https://github.com/pimcore/pimcore/blob/11.x/.github/workflows/php-cs-fixer.yaml and other repros

blankse: But yes I think we should set the latest major version. I will change it.

blankse: @fashxp Done https://github.com/pimcore/data-hub/pull/854/commits/759bb046266ce4c2f399156cffe86d4e5fef0e11

fashxp (approved):

fashxp:
@blankse hmm, it seems that this image does not exist? see https://github.com/pimcore/data-hub/actions/runs/8797410558

blankse:
@fashxp Maybe ":3" not work? Should we try it with ":latest"?

PR #860, opened 2024-04-29
PhpCsFixer: Execute on pull requests

blankse:
See https://github.com/pimcore/customer-data-framework/pull/532


fashxp:
thx very much!

Issue #865, opened 2024-05-27
[Improvement]: Fix inconsistency when creating/updating and running into a validation exception 


svdv22:
### Improvement description

For a project we hooked in onto the `DataObjectEvents::PRE_UPDATE` event to validate some data and we throw a `Pimcore\Model\Element\ValidationException` when the validation fails. This logic is executed every time we save a product.

When we use Datahub (GraphQL) to create/update objects, and validation fails, we get inconsistent responses. This is caused because we end up in different resolve functions.

In [MutationType.php#L839](https://github.com/pimcore/data-hub/blob/1.x/src/GraphQL/Mutation/MutationType.php#L839) you'll see that when saving fails it's being caught and you return an array with success and message.
However [around line 660](https://github.com/pimcore/data-hub/blob/1.x/src/GraphQL/Mutation/MutationType.php#L660) an eventual Exception is not caught. So you'll end up with a generic non-informative message (and stack trace in debug mode).

I think this could be easily fixed by adding a try/catch around line 660 too.

svdv22:
Since it's almost no effort I've created a PR already: https://github.com/pimcore/data-hub/pull/866

PR #866, opened 2024-05-27
Add try/catch to catch duplicate path + general exceptions

svdv22:
See https://github.com/pimcore/data-hub/issues/865


markus-moser:
Thanks a lot 👍

PR #869, opened 2024-07-04
Fix: Quantity Value "Warning: Undefined array key value"

smoothie:
**Context**
When a quantity value, defines no value it raises PHP exception undefined array key "value". 

**Fix**
Applies same behavior as for unit/unitId -> Use null as default

image


Corepex:
@smoothie, thanks for your PR. I tested it and everything runs perfectly. Great :100:  :1st_place_medal:

PR #875, opened 2024-08-20
[Bug] Mutations to tables in Bricks

### LINKED ISSUE #863: [Bug, EC]: Mutations to tables don't work at all
### Expected behavior

Mutations work.

### Discussed in https://github.com/orgs/pimcore/discussions/17026

<div type='discussions-op-text'>

<sup>Originally posted by **khenow1978** May  8, 2024</sup>
Hi,
I just came across an issue that really bothers me.

We want to use a table for storing data about quantity and measurement calculations. Therefore I'm currently testing how to query the content of a table and create new  entries. However as long as I use the table within the query schema all works fine but as soon as I add it to the Mutation schema the whole API crashes. This is the response: 
image

The documentation tells me that mutations of tables are supported in the current version.. we are running Pimcore 11.1.5. But somehow it doesn't work at all. Any suggestions?</div>

The problem occurs because the $tableDef variable is empty. This is not filled in the “getFieldDefinition”. This seems to be due to the name of the field:

Ueberstand~tblMengenumrechnung

The table is in an object brick and therefore there is the prefix “Ueberstand”. The function doesn’t seem to be able to cope with this and ends up with “return zero”.

I copied the element and used it directly in my class. This means that the GraphQL interface runs smoothly. I was able to work around/solve the problem for myself at first.

### Steps to reproduce

- Create an object type “table” in an object brick.
- If possible, this should have a fixed number of columns and headers.
- Add this object to a mutation.
- After that, the API should no longer run.

mcop1:
Resolves #863

### DIFF
--- a/src/Configuration.php
+++ b/src/Configuration.php
@@ -424,8 +424,6 @@ class Configuration extends AbstractModel
     /**
      * @internal
      *
-     * @param ?User $user
-     *
      * @return bool
      */
     public function isAllowed(string $type, ?User $user = null)
--- a/src/GraphQL/DataObjectInputProcessor/Table.php
+++ b/src/GraphQL/DataObjectInputProcessor/Table.php
@@ -19,6 +19,8 @@ use GraphQL\Type\Definition\ResolveInfo;
 use Pimcore\Bundle\DataHubBundle\GraphQL\Service;
 use Pimcore\Model\DataObject\Concrete;
 use Pimcore\Model\DataObject\Fieldcollection\Data\AbstractData;
+use Pimcore\Model\DataObject\Objectbrick\Definition;
+use RuntimeException;
 
 class Table extends Base
 {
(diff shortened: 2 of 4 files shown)


mcop1:
Hello @khenow1978 ,
I created this pull request after my attention was brought to your post in the data hub discussion. Could you please check if the changes in this pull request solve your problem?

khenow1978:
Hi.. I'm afraid I'm not a programmer and therefore I can't check the code. As soon as it has been released I can check it in the frontend but I can't really check it by looking at the code..

mcop1:
Got it, no worries, we test it internally anyways and release it as a bugfix as soon as possible. Just wanted to have another pair of eyes on it :)

blankse:
@mcop1 These changes have caused regressions. As a result, our schemas with block fields in object bricks no longer work.

PR #874, opened 2024-08-19
[Task]: Replace Request::get with explicit input sources 

kingjia90:
Resolves https://github.com/pimcore/data-hub/issues/679
Related https://github.com/pimcore/pimcore/issues/14138

PR #877, opened 2024-09-04
[Task]: Follow up to `Replace Request::get with explicit input sources`

mcop1:
Part of https://github.com/pimcore/data-hub/issues/679 , see https://github.com/pimcore/data-hub/pull/874

Despite being typed with `mixed`, `Array`can´t be passed as default value for `request->get`, needs to be a scalar value. The type gets checked in the function itself using `is_sclar` and therefore the type mismatch doesnt get detected by phpstan.

### DIFF
--- a/src/Controller/WebserviceController.php
+++ b/src/Controller/WebserviceController.php
@@ -93,6 +93,7 @@ class WebserviceController extends FrontendController
         LongRunningHelper $longRunningHelper
     ) {
         $clientname = $request->attributes->getString('clientname');
+        $variableValues = null;
 
         $configuration = Configuration::getByName($clientname);
         if (!$configuration || !$configuration->isActive()) {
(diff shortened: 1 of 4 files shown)


src/Controller/WebserviceController.php
```diff
@@ -178,7 +178,11 @@ public function webonyxAction(
             $this->eventDispatcher->dispatch($event, ExecutorEvents::PRE_EXECUTE);
 
             if ($event->getRequest() instanceof Request) {
-                $variableValues = $event->getRequest()->request->get('variables', $variableValues);
+                $variableValues = $event->getRequest()->request->get('variables');
+            }
+
+            if (empty($variableValues)) {
```
blankse: Variable is always set
```suggestion
            if (!$variableValues) {
```

PR #878, opened 2024-09-14
Fix that empty metadata is filled with the previous relation

skoch98:
If I update a metadata relation field via data-hub, all metadata fields that not provieded a value (or empty-value) in the promise will reviece the value from the last relation with "filled" metadata.

See:

Before (via Admin):
image

Promise:
image

After:
image


lukmzig:
@skoch98 thank you very much for the fix :+1:

PR #880, opened 2024-09-16
Use constant for  'datahub_context' instead hardcoded value

vmalyk:
Code improvements and using constant instead hardcoded key.


mcop1:
Looks good to me, thank you :+1:

PR #883, opened 2024-09-16
Added field type for table query type definition

### LINKED ISSUE #879: [Improvement]: Define DataObject table datatype more clear
### Improvement description

For some cases table and structure table require specific handling, but no way to clarify that , because  in generated typename - it contains parent type (object,field collection), class name, fieldname.
As suggestion, If adding additionally field type - it'll be helpful for improve field handling on integration side.

vmalyk:
Solves #879.


mcop1:
Thank you!

PR #889, opened 2024-09-26
perf: optimize the permission check to use an in-memory object

### LINKED ISSUE #886: [Improvement][Perf]: WorkspaceHelper::isAllowed creating hundreds of database queries for listings
### Improvement description

WorkspaceHelper::isAllowed will run a DB query for each item in the listing, including relations:
https://github.com/pimcore/data-hub/blob/f78112706afa25a590f3519d9ee9364df742ea2f/src/WorkspaceHelper.php#L280-L282

In my test, fetching 50 products and related images will run the query 426 times, accounting for about 30% of the total runtime.

Untitled-profile-Blackfire (1)

Looking at the code, it seems it should be possible to avoid these queries completely by creating a special in-memory (in cache) data structure optimized for this check and running the check as fast as possible on that specific structure.

Edit: when the listing is fully cached (all elements fully resolved from cache), the listing will still run 546 queries, but 426 of them are still this permission check, that's 80% of all the work done, accounting for 40% of total runtime.

dkarlovi:
Closes #886.

Comparison current and new, fetching 100 products in a listing:
Comparison-from-Untitled-to-Untitled-Blackfire-09-26-2024_03_04_PM

### DIFF
--- a/src/WorkspaceHelper.php
+++ b/src/WorkspaceHelper.php
@@ -20,9 +20,9 @@ use Pimcore\Bundle\DataHubBundle\Event\GraphQL\Model\PermissionEvent;
 use Pimcore\Bundle\DataHubBundle\Event\GraphQL\PermissionEvents;
 use Pimcore\Bundle\DataHubBundle\GraphQL\Exception\ClientSafeException;
 use Pimcore\Bundle\DataHubBundle\GraphQL\Exception\NotAllowedException;
+use Pimcore\Cache;
 use Pimcore\Cache\RuntimeCache;
 use Pimcore\Db;
-use Pimcore\Logger;
 use Pimcore\Model\DataObject\OwnerAwareFieldInterface;
 use Pimcore\Model\Element\ElementInterface;
 use Pimcore\Model\Element\Service;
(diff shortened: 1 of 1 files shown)


dkarlovi:

mattamon:
@dkarlovi tested it and it works fine. My only concern is that the `$cacheKey` does not get invalidated if you save the configuration. If the permissions change you need to call cache:clear to make it work. 
Could you also please invalidate the key on config save. 

Also maybe change the key to `datahub_workspace_permissions_` to make it more specific.

dkarlovi:
@mattamon is there a cache tag which gets cleared on workspace save, like `asset_ID` or `object_ID` do when they're saved? We can then add a tag to the cache item and it gets automagically invalidated.

dkarlovi:
@mattamon is there a cache tag which gets cleared on workspace save, like `asset_ID` or `object_ID` do when they're saved? We can then add a tag to the cache item and it gets automagically invalidated.

mattamon:
> @mattamon is there a cache tag which gets cleared on workspace save, like `asset_ID` or `object_ID` do when they're saved? We can then add a tag to the cache item and it gets automagically invalidated.

Hmm, I checked and it does not seem like it. The configs are in general not cached, other than by the locationaware dao. 
But this is a static cache. 

@fashxp do you have a tag in mind? Otherwise I would introduce a tag and change the Configuration Dao save method to clear that tag.

dkarlovi:
> The configs are in general not cached

IMO that's also an opportunity to improve: why aren't they cached, wouldn't the DAO do that by default? :thinking:  This would mean we'd get cache invalidation for free here just by tagging.

mattamon:
> > The configs are in general not cached
> 
> IMO that's also an opportunity to improve: why aren't they cached, wouldn't the DAO do that by default? 🤔 This would mean we'd get cache invalidation for free here just by tagging.

Yeah I thought the same.
Like I wrote they are kind of cached here https://github.com/pimcore/pimcore/blob/11.x/lib/Model/Dao/PimcoreLocationAwareConfigDao.php but in a static cache.

fashxp:
> @fashxp do you have a tag in mind? Otherwise I would introduce a tag and change the Configuration Dao save method to clear that tag.

nope, no tag in mind... let's introduce a new one.

dkarlovi:
@mattamon friendly ping, did you maybe get a chance to add the cache tag flush on save by any chance in the meantime?

mattamon:
@dkarlovi ahhh, I actually thought that you will introduce it, since it can be done in the same scope within your PR.
Sorry maybe there was some miscommunication on my part cause I also asked to change `datahub_workspace_permissions_` and for me it was clear that you need to do that update on the PR.

So I guess, I have to ask now, can you please introduce this tag in your PR? Would be awesome!

dkarlovi:
> I would introduce a tag and change the Configuration Dao save method to clear that tag.

Yes, this directly told me you'll be creating a separate PR to do this. :) OK, I'll try to do both here.

mattamon:
> > I would introduce a tag and change the Configuration Dao save method to clear that tag.
> 
> Yes, this directly told me you'll be creating a separate PR to do this. :) OK, I'll try to do both here.

Thank you! And sorry again for the confusion!

dkarlovi:
@mattamon ready for review, should be OK now.

mattamon (approved):
Thank you very much! It works perfectly now!

PR #890, opened 2024-09-26
Fix regression from #875

blankse:
Fix regression from https://github.com/pimcore/data-hub/pull/875

After this no block field in object bricks doesn't work.


blankse:
@mcop1 I create a PR for the regression. Can you review it?

mcop1:
Thanks you very much for fixing the regression.

PR #900, opened 2024-10-22
Use factory for Pimcore object type instead static call.

vmalyk:
Hi,

It looks like Document and Asset have factories that generate QueryType and provide the ability to replace and extend entity schema definitions more easily. However, this does not work for Data Objects.
I suggest adding a similar approach for Data Objects with its own factory, as we need to pass parameters, context, and class names without breaking the current implementation.

Please review.

### DIFF
--- a/src/GraphQL/ClassTypeDefinitions.php
+++ b/src/GraphQL/ClassTypeDefinitions.php
@@ -37,12 +37,8 @@ class ClassTypeDefinitions
     {
         $db = Db::get();
         $listing = $db->fetchAllAssociative('SELECT id, name FROM classes');
-
         foreach ($listing as $class) {
-            $id = $class['id'];
-            $name = $class['name'];
-            $objectType = new PimcoreObjectType($graphQlService, $name, $id, [], $context);
-            self::$definitions[$name] = $objectType;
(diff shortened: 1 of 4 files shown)


mcop1 (commented):
Hello,

thank you for opening the pr, the changes look really good. 

Just a few small things:

- I rebased the pr to the 1.x branch, we think it would fit better in a minor version as improvement rather than in a bugfix release
- For new classes it is ok to use native types, e.g. `protected string $className` instead of `protected $className`. Could you please check and add the proper types? 
- I guess you probably just copy/pasted that part of the code, but is there a use case where we would need to extend the `DataObjectTypeFactory`? If not, t hen I would suggest to make the class final and set the `$className` property to private. 

Thank you for your efforts.

vmalyk:
Hi @mcop1,

thank you for feedback and I updated  factory by using native types. I think it's better leave methods as protected and leave possibility to inherit new classes from factory and leave factory in the consistency with previously added factories.

Please have a look.

mcop1:
Hello @vmalyk 
agreed, we can keep it consistent. I created a ticket to take a look at the classes and make them final/internal (if it is fitting) for the next major release.

Thank you again for your efforts!

PR #910, opened 2024-11-06
Fix regression from #875

blankse:
Followup to #875 and #890

$brickDescriptor is an array. I get following errors:
```
Uncaught TypeError: Pimcore\Bundle\DataHubBundle\GraphQL\Service::getValueFromObjectBrick(): Argument #4 ($brickDescriptor) must be of type ?string, array given
```


blankse:
Please fix this also with a 1.7.5 release. The 1.7.3 and 1.7.4 releases are buggy and 1.8.x can't be installed in Pimcore 10.

mcop1:
@blankse  Thanks again, I will do a 1.7.5 bugfix release, merge it to 1.8 afterwards and do a bugfix release there too.

PR #912, opened 2024-11-13
Added CsFeature title

dturopoli:
Added title to classification store key graphql output.

Should we also consider admin translations for titles?


mcop1:
Hello @dturopoli ,

pr looks good to me, thank you. I would like to have it in the next release, so I would merge it right now, that way we provide the basic functionality. Adding admin translations would be nice, but that can still be done in a follow up.

Thanks!

PR #916, opened 2024-11-22
Allow cloning even if record is not writeable

cancan101:
The ability to clone does not depends on the source record being  writeable.


mcop1:
Hello @cancan101 ,
agreed, thanks for the pr.

PR #928, opened 2025-01-15
Improve GraphQL caching for large queries

gajdusek:
Pimcore limits caching to a maximum of 50 items per request. For large GraphQL queries, this limit can easily be exceeded because every object loaded during query resolution is added to the cache queue with priority 0. When the queue is truncated, only the top 50 items (by priority) are retained.

This commit increases the priority of the final GraphQL query result, ensuring this is more likely to be cached despite the 50-item limit.


robertSt7:
@gajdusek Could you please resolve the conflicts? Thanks

fashxp:
can't be merged anymore

PR #939, opened 2025-02-19
[Improvement] Add possibility to retrieve specific versions

### LINKED ISSUE #546: [Feature]: GraphQL, get specific versions
### Feature description

GraphQL queries are only able to return the currently published version of individual documents/assets/data-objects. For certain usecases it would be useful to also access specific unpublished versions/drafts, especially in a headless frontend context. One such example would be to render previews of newly added/changed headless-documents before publishing.

We propose to add an optional version-id argument to the `getDocument()`, `getAsset()` and generated data-object getter functions within GraphQL queries.

Ideally this feature should be activable/deactivatable within the security definitions of the datahub configuration using a new checkbox.

schulzjona:
Resolves https://github.com/pimcore/data-hub/issues/546

This adds the possibility to add a version parameter to the GraphQL Query to retrieve specific versions of:
- Assets
- Objects
- Documents


schulzjona:

robertSt7:
@schulzjona With Assets this works fine for me, but with Documents and DataObjects I always get the data from the latest version. Could you please take a look at this? Thanks

sm00thelf:
👍

sm00thelf:
image
@robertSt7 Can u please check if this is checked in your Security Definition? When it is checked it always returns latest versions data

robertSt7:
> image @robertSt7 Can u please check if this is checked in your Security Definition? When it is checked it always returns latest versions data

@schulzjona This isn't checked in my config. Can you please try to query the `modificationDate` of an object and see if it works for you? Thanks

src/Configuration.php
```diff
@@ -206,6 +206,11 @@ public function skipPermisssionCheck()
         return $this->configuration['security']['skipPermissionCheck'] ?? false;
     }
 
+    public function disableVersionedRequests(): bool|string
+    {
+        return $this->configuration['security']['disableVersionedRequests'] ?? false;
```
cancan101, line 211: for safety reasons, shouldn't this default to true?

sm00thelf: Yes, that is probably right!

sm00thelf:
Had some pretty busy weeks. Looking forward to continue work on this during this week/weekend 👍

sm00thelf:
Closing this for now as I'm currently unable to dedicate the necessary time. I might revisit and rework this in the future, but until then, I don't want to unnecessarily clutter the PR list.

PR #944, opened 2025-02-24
Add concept of IsValidDataObjectTriggerEvent.

laticrete-pimcore:
Carried over from: https://github.com/pimcore/data-hub-file-export/pull/181

CC @fashxp

### DIFF
--- a/src/Event/IsValidDataObjectTriggerEvent.php
+++ b/src/Event/IsValidDataObjectTriggerEvent.php
@@ -1,0 +1,48 @@
+<?php
+
+/**
+ * Pimcore
+ *
+ * This source file is available under two different licenses:
+ * - GNU General Public License version 3 (GPLv3)
+ * - Pimcore Commercial License (PCL)
+ * Full copyright and license information is available in
+ * LICENSE.md which is distributed with this source code.
+ *
+ *  @copyright  Copyright (c) Pimcore GmbH (http://www.pimcore.org)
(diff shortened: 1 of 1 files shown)


src/Event/IsValidDataObjectTriggerEvent.php
```diff
@@ -0,0 +1,48 @@
...
+use Pimcore\Event\Model\ElementEventInterface;
+use Symfony\Contracts\EventDispatcher\Event;
+
+class IsValidDataObjectTriggerEvent extends Event
+{
+    protected bool $isValid;
+
+    protected ElementEventInterface $event;
+
+    public function getEvent(): ElementEventInterface
```
fashxp: should we name that something like `sourceEvent`?

fashxp (changes requested):

src/Event/IsValidDataObjectTriggerEvent.php
cancan101: how about `triggerEvent` for consistency with the name of the class?

fashxp: how should we proceed here?

laticrete-pimcore: if you are good with the suggested name (I see your thumbs up), I will update the PR.

fashxp:
:heart:  thx very much

PR #946, opened 2025-02-25
[Improvement] Final/internal classes

### LINKED ISSUE #906: [Improvement]:  Check internal/final classes
### Improvement description

Check classes/interfaces/... in data-importer and decide if they should be internal and/or final.
Provides better flexibility in the future.

mcop1:
Resolves #906 

## Additional information
Using the final keyword distinguish data-hub internal API from the external API intended for developers. 
By marking classes as final, we prevent unintended inheritance, ensuring internal logic remains stable while allowing extensions through interfaces, abstract classes and/or composition. 
This approach reduces the risk of breaking changes when modifying internal classes, as external developers cannot extend them directly. Instead, we try to maintain flexibility through explicitly provided extension points, leading to a more maintainable implementation.


src/GraphQL/DataObjectMutationFieldConfigGenerator/Base.php
```diff
@@ -21,6 +21,9 @@
 use Pimcore\Bundle\DataHubBundle\GraphQL\Traits\ServiceTrait;
 use Pimcore\Model\DataObject\ClassDefinition;
 
+/**
```
markus-moser: I think this one should not be internal. Can be useful when someone adds it's own datatypes. Yes we also have the interface, but let's also make the base class public. WDYT?

src/GraphQL/Resolver/QueryType.php
```diff
@@ -33,7 +33,10 @@
 use Pimcore\Model\Translation;
 use Symfony\Component\EventDispatcher\EventDispatcherInterface;
 
-class QueryType
+/**
```
markus-moser: This should be public as it is used in `QueryTypeEvent`

src/GraphQL/SharedType/JsonType.php
```diff
@@ -18,7 +18,10 @@
 use GraphQL\Type\Definition\ScalarType;
 use GraphQL\Utils\Utils as GraphQLUtils;
 
-class JsonType extends ScalarType
+/**
```
markus-moser: Let's make this public is this sometimes can be useful for custom operators.

markus-moser (changes requested):
In general it LGTM 👍 

Just found a few places to check

src/GraphQL/DataObjectMutationFieldConfigGenerator/Base.php
mcop1: Agreed, makes sense.

mcop1:
> In general it LGTM 👍
> 
> Just found a few places to check

Thanks for the review, I adapted the mentioned files.

markus-moser (approved):

PR #948, opened 2025-03-03
Allow setting Date and DateTime-fields to empty / null even if the fieldtype is DATETIME

ferdinand-it:
Currently, Date or DateTime fields to which zero or an empty string is passed behave differently depending on whether BIGINT or DATETIME is specified as the FieldType. For the latter, it is not possible to set the date empty via graphql. Instead, the date is set to “1970-01-01 01:00:00” in this case.

grafik

To solve this issue, it is now checked if the value pasased via graphql is an empty string or null.


ferdinand-it:
recheck

ferdinand-it:
recheck

PR #950, opened 2025-03-04
Allow setting Date and DateTime-fields to empty / null even if the fieldtype is DATETIME

ferdinand-it:
Currently, Date or DateTime fields to which zero or an empty string is passed behave differently depending on whether BIGINT or DATETIME is specified as the FieldType. For the latter, it is not possible to set the date empty via graphql. Instead, the date is set to “1970-01-01 01:00:00” in this case.

grafik

To solve this issue, it is now checked if the value pasased via graphql is an empty string or null.

### DIFF
--- a/src/GraphQL/DataObjectInputProcessor/Date.php
+++ b/src/GraphQL/DataObjectInputProcessor/Date.php
@@ -35,10 +35,17 @@ class Date extends Base
     {
         $attribute = $this->getAttribute();
         Service::setValue($object, $attribute, function ($container, $setter) use ($newValue) {
-            if (!is_numeric($newValue)) {
-                $newValue = strtotime($newValue);
+
+            if ($newValue === '') {
+                $newValue = null;
             }
-            $newValue = Carbon::createFromTimestamp($newValue);
+
+            if (!is_null($newValue)) {
+                if (!is_numeric($newValue)) {
+                    $newValue = strtotime($newValue);
+                }
+                $newValue = Carbon::createFromTimestamp($newValue);
+            }
 
             return $container->$setter($newValue);
         });


ferdinand-it:
recheck

ferdinand-it:
recheck

ferdinand-it:
@brusch @fashxp do you have any idea what I'm doing wrong with the CLA check? 
I signed the CLA and used my primary github mail address for the commit. But the check still fails...
grafik

ferdinand-it:
recheck

ferdinand-it:
recheck

fashxp:
thx for your contribution. 
concerning CLA check ... does on of the e-mail addresses configured in your github account match the e-mail address in the commit of that PR?

ferdinand-it:
> thx for your contribution. concerning CLA check ... does on of the e-mail addresses configured in your github account match the e-mail address in the commit of that PR?

Thanks for your reply. Yes, it does. I also retried the commit with the Primary Address in github, but it didn't work.
Now I unchecked the "Keep my email addresses private"-box. I could try to create a new commit/PR with this setting. If this won't work, I have now more ideas... But there are no general issues with the CLA-check and I'm the only one with such problems?

fashxp:
when you open https://github.com/pimcore/data-hub/pull/950.patch, you see the e-mail address of the commit. 

I'm not aware of any general issues with the CLA. Most of the time the e-mail address mismatch is the reason for failing...

fashxp:
another thing you could try - if not done already - is to create the directly in the web-ui of github.

ferdinand-it:
ok, thanks. The mailaddress in the commit is correct. I'll try the web-ui.

blankse:
@ferdinand-it I think you only need to create a pull request comment here with following content:

`I have read the CLA Document and I hereby sign the CLA`

ferdinand-it:

fashxp:
Ah damn, I missed the obvious 😜
Thx @blankse

ferdinand-it:
> @ferdinand-it I think you only need to create a pull request comment here with following content:
> 
> `I have read the CLA Document and I hereby sign the CLA`

Great, looks like this worked.

fashxp:
thx yery much

PR #953, opened 2025-03-10
[Fix]: advancedManyToManyRelation input metadata fix

dturopoli:
When adding multiple elements with metadata to advancedManyToManyObjectRelation 
$columns array was never cleared which resulted in failed request, since $columns would also contain metadata from previous iterations.


robertSt7:
@dturopoli As this is a bugfix, could you please rebase your PR to 1.9? Thanks

fashxp:
let's check and merge it to 2.1

robertSt7:
@dturopoli Thanks for fixing the metadata

PR #967, opened 2025-04-11
Drop legacy check for Symfony 4 and use constants in return

vmalyk:
hi!

Just small fix and clean up commands. Removed checking  constant and a bit adjust how it's used - it's public constant and might be no require import \Symfony\Component\Console\Command\Command. 

Additionally, removed null from return type in doctypes.  

Please review,


astapc:
Hello @vmalyk - thank you very much for the pull request.
Since it is based on an older version, we will not merge your PR to version 1.9.
However, the issue has already been resolved in version 2.3 through a different pull request.

PR #969, opened 2025-04-11
[fix] wrong init $rootValue value

pierre-kubica:
This $rootValue var is initiated with an empty array, I don't know why.
That variable has null default value in all function definition using it.

In my case an error occured using api-platform/core bundle [api-platform/core](https://github.com/api-platform/core) which uses webonyx/graphql-php bundle (also used by pimcore-datahub bundle). And api-platform correctly assumes that $rootValue will be null by default.


pierre-kubica:
recheck

PR #976, opened 2025-05-06
[fix] no 'update' perm needed for 'create' dataobject operation

pierre-kubica:
Here is a clean PR from my previous here (cause I messed up that one): #969

Hope it'll work this time :D


robertSt7:
@pierre-kubica Thx for opening a new PR and fixing the permission :)

PR #986, opened 2025-08-07
[Bug]: Unable to Clone Read-Only Configs

### LINKED ISSUE #958: [Bug]: Unable to Clone Read-Only Configs
### Expected behavior

Even if a config is read only, it should be able to be cloned.

### Actual behavior

Error:
>Error cloning configuration:

### Steps to reproduce

Create a config that read only (e.g. stored as config file).
This code looks wrong:
https://github.com/pimcore/data-hub/blob/19b93a13d843a1a850b7166df0309596149a9cd6/src/Controller/ConfigController.php#L211-L213

mcop1:
Resolves #958

### DIFF
--- a/src/Controller/ConfigController.php
+++ b/src/Controller/ConfigController.php
@@ -208,9 +208,6 @@ class ConfigController extends \Pimcore\Controller\UserAwareController
             if (!$originalConfig) {
                 throw new \Exception('Configuration not found');
             }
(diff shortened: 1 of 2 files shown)


mcop1:
Cherry picked to 2.1 in d44207723b54656c32f865ca37701f0fb7f53a1f

PR #987, opened 2025-08-15
Added system field "key" as optional argument for data object mutation

### LINKED ISSUE #988: [Improvement]: Treat system field "key" the same way in create and update of data objects
### Improvement description

The field "key" is a system field and should be treated the same way during both creation and updating. While it's possible to add the "key" field to the mutation configuration, this would mean that during creation, "key" would exist both as a parameter and a field on the input object, which can be confusing. In the update case, it would only be available as a field on the input object. Since "key" is required during creation, the parameter must be kept.

SimonCremer:
The field "key" is a system field and should be treated the same way during both creation and updating. While it's possible to add the "key" field to the mutation configuration, this would mean that during creation, "key" would exist both as a parameter and a field on the input object, which can be confusing. In the update case, it would only be available as a field on the input object. Since "key" is required during creation, the parameter must be kept.

### DIFF
--- a/src/GraphQL/Mutation/MutationType.php
+++ b/src/GraphQL/Mutation/MutationType.php
@@ -726,6 +726,7 @@ final class MutationType extends ObjectType
                     'type' => $updateResultType,
                     'args' => [
                         'id' => ['type' => Type::int()],
+                        'key' => ['type' => Type::string()],
                         'fullpath' => ['type' => Type::string()],
                         'parentId' => ['type' => Type::int()],
                         'defaultLanguage' => ['type' => Type::string()],
@@ -880,6 +881,16 @@ final class MutationType extends ObjectType
 
                 if (isset($args['omitMandatoryCheck'])) {
                     $object->setOmitMandatoryCheck($args['omitMandatoryCheck']);
+                }
+
+                if (isset($args['key'])) {
+                    if (!DataObject\Service::isValidKey($args['key'], AbstractObject::OBJECT_TYPE_OBJECT)) {
+                        return [
+                            'success' => false,
+                            'message' => '"key" is not valid',
+                        ];
+                    }
+                    $object->setKey($args['key']);
                 }
 
                 $tags = [];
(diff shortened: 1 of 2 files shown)


SimonCremer:

SimonCremer:
@fashxp Are there any plans if (and when) the change will be merged?

fashxp:
It is on our list for the next platform version end of this month.

src/GraphQL/Mutation/MutationType.php
```diff
@@ -882,6 +883,16 @@ public function getUpdateObjectResolver($processors, $localeService, $object = n
...
                 }
 
+                if (isset($args['key'])) {
+                    if (!DataObject\Service::isValidKey($args['key'], AbstractObject::OBJECT_TYPE_OBJECT)) {
+                        return [
+                            'success' => false,
+                            'message' => '"key" is not valid',
+                        ];
+                    }
+                    $object->setKey($args['key']);
```
kingjia90, line 893: Thank you for your PR!

LGTM so far, just few inputs.

When trying to update to a key that happens to be already existing, maybe a  more clearer message that the update failed due to trying to update to an existing key?

similar to https://github.com/pimcore/data-hub/blob/95bce3895a5aa6bf2c15e8c9a5e0ff2282146e6c/src/GraphQL/Mutation/MutationType.php#L662-L666
but for update

and an entry in the changelog would be good to announce this change
https://github.com/pimcore/data-hub/blob/2.x/doc/01_Installation_and_Upgrade/01_Upgrade_Notes.md

kingjia90: Please also merge the latests `2.x` branch (with https://github.com/pimcore/data-hub/pull/990) into this to re-run the tests, it should fix the issue about the secret missing

SimonCremer: @kingjia90 
I've added an entry in the changelog and merged the latest 2.x changes. 

Introducing a dedicated catch phrase for the `DuplicateFullPathException` looks like duplicate code for me. The Exception is already catched in the generic catch phrase with a message like `Duplicate full path [ REAL_FULL_PATH ] - cannot save object`.

kingjia90: ah ok, it should be just fine as is then.

Thank you again for the PR, i am merging it

PR #1001, opened 2025-11-07
DocumentFolderType: Missing type and properties fields

blankse:
All document types have type and properties fields, but these are missing for folders.


blankse:
@astapc Can you review this?

blankse:
@astapc Can you review this please? FYI @kingjia90

src/GraphQL/DocumentType/DocumentFolderType.php
```diff
@@ -58,6 +59,17 @@ public function build(&$config)
                     'type' => Type::string(),
                     'resolve' => [$resolver, 'resolveModificationDate'],
                 ],
+                'type' => Type::string(),
```
kingjia90, line 62:
```suggestion
```

this is a mistake from copy&paste, or? there is no wrapper where this belongs to

blankse: @kingjia90 Not a copy&paste mistake. `type` is the field name here, using graphql-php's shorthand where a
field can map directly to a Type instance. Same shorthand is used right above for 'filename' => Type::string()

blankse: @kingjia90 I think `filename` was a copy&paste mistake. I changed it to `key`. https://github.com/pimcore/data-hub/pull/1001/commits/245e3924e3c9f70b2e57ab905d3861a24831a74e

src/GraphQL/DocumentType/DocumentFolderType.php
```diff
@@ -58,6 +59,18 @@ public function build(&$config)
...
+                'properties' => [
+                    'type' => Type::listOf($propertyType),
+                    'args' => [
+                        'keys' => [
+                            'type' => Type::listOf(Type::string()),
+                           'description' => 'List of property key names to include '
```
blankse:
```suggestion
                            'description' => 'List of property key names to include '
```

PR #1004, opened 2025-11-18
AreablockDataType: Remove _editableType and _editableName fields

blankse:
Removed the '_editableType' and '_editableName' fields from the AreablockDataType configuration.

If you use these fields, an internal error occurs.

`Call to a member function getType() on array`
`Call to a member function getName() on array`


blankse:
@kingjia90 Can you also review this PR please?

src/GraphQL/DocumentElementType/AreablockDataType.php
```diff
@@ -33,22 +33,6 @@ public function __construct(Service $graphQlService)
             [
                 'name' => 'document_editableAreablock_data',
                 'fields' => [
-                    '_editableType' => [
-                        'type' => Type::string(),
-                        'resolve' => static function ($value = null, $args = [], $context = [], ?ResolveInfo $resolveInfo = null) {
-                            if ($value) {
-                                return $value->getType();
```
kingjia90, line 40: not sure about the removal, what about`$value['type']` instead?

blankse: It already exists a type field

blankse: And as far as I can remember what happened in November, there was no ‘name’ key in the array :D

PR #1005, opened 2025-11-18
Add missing DocumentElementType Renderlet

blankse:
Adds GraphQL support for the Renderlet document editable, which was previously missing from the document element types.

Introduces a new RenderletType exposing the editable's _editableType, _editableName, id, type, subtype and the referenced relation (resolved as an anytarget element), plus the corresponding Renderlet query field config generator and its service registration.


blankse:
@kingjia90 Can you review this also please?

blankse:
@kingjia90 Every time I work with the DataHub GraphQL API, I find errors. It’s happened again today https://github.com/pimcore/data-hub/pull/1115. It seems I’m the only one submitting bug fixes. But these have been left unresolved since November. I wonder if we’re the only ones using this API...

kingjia90:
Solid structure (mirrors `RelationType`, wiring is correct), but two real bugs and some noise:

1. **`RenderletType.php:77` — `relation` is almost always `null` in production.** `getO()` doesn't lazy-load, and `$o` is stripped when documents are serialized into the core cache. Any cache-hit document resolves `relation` to `null` despite a valid target. Fix: call `$value->load()` first (like core does everywhere).
2. **`RenderletType.php:77` — leaks unpublished targets.** `Relation::getElement()` filters unpublished elements; `getO()` doesn't, and nothing downstream compensates. Add a publish check alongside the `load()` fix.

3. **`RenderletType.php:59` — `type` field is useless.** `Renderlet::getType()` is hardcoded to `'renderlet'`, so `type` always equals `_editableType` and never tells you if the target is an asset/document/object. Use the target's real type (what `getData()['type']` returns) — it's new public schema, hard to change later.

5. **Docs** — `01_Document_Queries.md` supported-editables list not updated.

blankse:
Fixed in 980e713c.

1. **`relation` null on cache hits** — now calls `$value->load()` before `getO()`. `getO()` doesn't lazy-load and `$o` is stripped by `Renderlet::__sleep()`, so cache-hit documents were resolving to `null`.
2. **Unpublished targets** — added a `Element\Service::doHideUnpublished()` / `isPublished()` filter, mirroring what `Relation::getElement()` does internally.
3. **`type` field** — now resolves via `getData()['type']` (the real asset/document/object type) instead of `Renderlet::getType()`, which is hardcoded to `'renderlet'` and duplicated `_editableType`.
5. **Docs** — added Renderlet to the supported page element types list in `01_Document_Queries.md`.

PR #1023, opened 2026-01-29
Added system field "published" as optional argument for data object mutation

AlwaysTiredOstrich:
Very similar to https://github.com/pimcore/data-hub/pull/987

The system field `published` should be treated the same way during both creation and updating. Adding this field to the mutation configuration poses the same issues as described for the field `key` in the linked PR.


AlwaysTiredOstrich:

SimonCremer:
Is there a chance that this will be merged?

kingjia90:
Thank you for your PR!
LGTM, just need to add a changelog, and it might need to wait for next minor release to be added, other than that can be merged

SimonCremer:
@kingjia90 I've updated the PR and the Upgrade Notes. I hope this was correct?

PR #1045, opened 2026-02-19
Fix GraphQL element date resolver type error

umizoomi:
## Root Cause

The `resolveModificationDate()` and `resolveCreationDate()` methods were added in DataHub v2.x with strict `array` type hints, but they don't account for `ElementDescriptor` objects that are returned when resolving relation fields.

## Affected Queries

```graphql
getObject {
  relation {
    creationDate
    modificationDate
  }
}
```

## Related Files
```
src/GraphQL/RelationHelper.php
└── processRelation() returns ElementDescriptor

src/GraphQL/ElementDescriptor.php
└── extends ArrayObject (array-accessible but not type-compatible)

src/GraphQL/BaseDescriptor.php
└── extends ArrayObject
```


umizoomi:

umizoomi:
recheck

PR #1134, opened 2026-07-27
Fix frontend workflows checkout for private-fork PRs (2.4)

bluvulture:
## Problem

`shared-frontend-build.yaml` and `shared-npm-publish.yaml` check out the PR head **directly from the fork** (`repository: ${{ github.event.pull_request.head.repo.full_name }}`). Private forks (partner orgs, e.g. `laticrete-pimcore/data-hub`) are not readable by this repo's `GITHUB_TOKEN`, so the fetch fails with `remote: Repository not found.` and every job dies at Checkout. `frontend-build-pr.yaml` triggers on `pull_request_target`, so every private-fork PR touching the frontend hits this.

Same failure already observed on data-hub-simple-rest: [run 30102786641](https://github.com/pimcore/data-hub-simple-rest/actions/runs/30102786641/job/89513568796?pr=310); same fix as [workflows-collection-public#142](https://github.com/pimcore/workflows-collection-public/pull/142).

## Fix

- New `CHECKOUT_REF` env: fork PRs fetch `refs/pull/<n>/head` from the **base repo** (mirrored there for every PR regardless of fork visibility); same-repo PRs and push/tag runs keep the branch/tag tip exactly as before. The `repository:` override is dropped.
- The two `git-auto-commit` steps (lint fixes, build output) are gated to same-repo contexts: pushing to a fork branch was never possible with this repo's token, and on the detached `refs/pull/<n>/head` checkout the action would fail rather than skip.

## Notes for reviewers

- `shared-npm-publish.yaml` is currently only called from push/tag triggers, where `head.repo` is empty and checkout already fell back to the base repo — fixed anyway for consistency/future-proofing.
- **Security follow-up (out of scope here):** under `pull_request_target`, the lint/check-types/build jobs run fork-controlled code (`npm ci` hooks, npm scripts) while workflow-level `permissions: contents: write` applies and checkout persists the token in `.git/config`. The upstream reusable workflow (`reusable-studio-frontend-build.yaml`) solves this properly (per-job permissions, `persist-credentials: false`, patch-artifact commit flow) — consider migrating these local copies to it.

## Verification

- YAML validated with js-yaml.
- push / dispatch → `CHECKOUT_REF` = `github.ref_name` (unchanged); same-repo PR → `github.head_ref` (unchanged); fork PR → `refs/pull/<n>/head` from base repo (fixes the failure).

### DIFF
--- a/.github/workflows/shared-frontend-build.yaml
+++ b/.github/workflows/shared-frontend-build.yaml
@@ -3,6 +3,14 @@ name: "Shared Frontend Build"
 on:
   workflow_call:
 
+env:
+  # Ref for all checkouts. Fork PRs: the fork can be private (partner orgs),
+  # and this repo's GITHUB_TOKEN cannot read it — fetching the fork directly
+  # fails with "Repository not found". Fetch the PR head as mirrored into the
+  # BASE repo (refs/pull/<n>/head) instead; that ref exists for every PR
+  # regardless of fork visibility. Same-repo runs keep the branch tip.
+  CHECKOUT_REF: ${{ github.event.pull_request && github.event.pull_request.head.repo.full_name != github.repository && format('refs/pull/{0}/head', github.event.pull_request.number) || github.head_ref || github.ref_name }}
+
 permissions:
   contents: write
 
@@ -14,8 +22,7 @@ jobs:
       - name: Checkout code
         uses: actions/checkout@v4
         with:
-          ref: ${{ github.head_ref || github.ref_name }}
-          repository: ${{ github.event.pull_request.head.repo.full_name }}
+          ref: ${{ env.CHECKOUT_REF }}
 
       - name: Set up Node.js
         uses: actions/setup-node@v4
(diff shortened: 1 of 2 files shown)


berfinyuksel:
Cross-referencing from the workflow-parity work (product-management#1305):

**The fix looks right.** `refs/pull/<n>/head` from the base repo is the correct pattern — it exists for every PR regardless of fork visibility, drops the cross-repo `repository:` override, and the same-repo gating of the two `git-auto-commit` steps matches what that action can actually do. Verified the `CHECKOUT_REF` expression resolves correctly in `workflow_call` context (caller's event is inherited) and keeps push/tag runs on the branch tip exactly as before.

Two related notes:

1. The residual risk your "Security follow-up" paragraph names is now formally tracked: these two files plus `frontend-build-pr.yaml` (here and on copilot-bundle 2.5) were added to the pull_request_target review scope in service-operations#1083 (2026-08-14 addendum) — fork-controlled `npm ci`/scripts under prt with `contents: write` and the npm token in the publish path. This PR neither worsens nor resolves that; correct to keep it out of scope here.
2. **For whoever merges:** these files exist only on the 2.4 line (2026.x runs the unified pipeline). The post-merge forward-merge `2.4 → 2026.1 → 2026.2 → 2026.x` must **not** carry them upward as clean adds — the sync tooling from #1305 now drops workflow files absent on 2026.x during chain merges, so running that routine (or dropping the three files manually at each leg) keeps the legacy prt pipeline off the newer lines.

PR #1150, opened 2026-09-15
One summary for both of a configuration's stories

markus-moser:
Every Data Hub adapter that shows an agent's proposal about a configuration draws the same rail: the configuration it is about, then a dotted spine with a node per section, a section label that opens it in the editor, and a mark per changed field. It existed twice — in `pimcore/data-importer` and `pimcore/data-hub-webhooks`, 388 lines byte-identical — with three more adapters to come. It moves here.

## The interface

```ts
interface ConfigSummaryRow     { key, label, note?, hint?, status? }
interface ConfigSummarySection { key, label, rows: ConfigSummaryRow[] }
interface ConfigSummaryFoot    { lead, detail? }
type      ConfigSummaryVariant = 'changes' | 'description'

interface ConfigSummaryProps {
  name, description?, active?
  sections, variant?, foot?
  activeKey?, onOpenSection?
}
```

Data, not slots. Both of a configuration's stories turn out to be the same shape — a primary line, an optional second line, an optional mark. Describing a change the primary line is the field's name and the mark says what became of it (`changes`); describing a configuration that does not exist yet the primary line is the value and there is no mark (`description`). One component, one `variant`, no `children` escape hatch needed.

The foot is two plain strings — `lead` may mark its own emphasis with `**…**`, so a translator decides what carries it. The changed-field count is derived from `sections`; the four status words, the counts, the on/off label and "New" ship here in all seven Studio locales.

## Theme

Styles come through `createStyles` from `@pimcore/studio-ui-bundle/app`, as Studio's own components do. Imported from `antd-style` directly, a plugin resolves that package's own copy — outside Studio's `ThemeProvider`, so every `token.color*` is antd's stock palette and the primary renders `#1677ff` beside Studio's violet. Measured in the same modal: the section labels and spine nodes now match Studio's own switch colour exactly.

Every other style sheet in this bundle still imports `antd-style` directly and has the same problem; that is a separate cleanup.

The spine node is positioned off the heading's line box — `(fontSizeSM × lineHeightSM − 7) / 2` — rather than a fixed margin, and the heading is block-level rather than inline, so the node stays centred on its headline whatever the theme makes of the small font.

## Why this bundle and not Change Control

Presentation only. It knows nothing about how a change is stored, fetched or approved. The other duplicated files — the review payload types, the change-set fetch hook, the form-annotation probe — carry Change Control's wire contract; this bundle has no Change Control knowledge, and moving that here would let a Change Control release break Data Hub while still leaving Change Control's own subjects unable to use it. That half belongs in a published review-contract package and stays duplicated until there is one.

## Gates

`tsc --noEmit`, eslint, `rsbuild build` and `generate-types` green. Both consumers build against this source and render both variants; the Importer's rail and brief, and the Webhooks rail, are migrated in their own PRs and merge after this releases a canary.


markus-moser:
Superseded: the summary moves into Change Control's review lane, next to the review contract it belongs to — pimcore/enterprise-monorepo#95 (Studio SDK) and https://github.com/pimcore/enterprise-monorepo/pull/96. Consumers switch their import to `@pimcore/change-control-bundle/sdk`.

